PR #16, opened 2020-09-24
fix: spawning process

### LINKED ISSUE #15: melos exec hang
when i run `melos exec -c=1 -- flutter pub run build_runner build --delete-conflicting-outputs` it hang after the completion of the first package.

from what i have debug it seems `await processStdoutCompleter.future` line [218](https://github.com/invertase/melos/blob/master/lib/src/common/utils.dart#L218) never complete but i didn't understand why for now :/

EDIT: After more investigation it seems the exit command pushed in stdin after the command is eat by the running program.

EDIT2: Also i'm not sure why you are starting a shell yourself since the `runInShell` parameter on `Process.start` already do it for you ?

Kleak:
fix #15 

tested on macos

⚠️⚠️⚠️
please try it on a windows laptop before merging since i don't have one.
thanks
⚠️⚠️⚠️


Salakar:
> Also i'm not sure why you are starting a shell yourself since the runInShell parameter on Process.start already do it for you ?

There was a reason for this, can't quite remember now what it was. I'll look into this, need to test it on Windows, but the code in this PR looks like what it was originally which had some issues. Will get back to you. Thanks for the PR :)

ened:
@Kleak could you please rebase your branch

PR #18, opened 2020-10-03
Fix for newer dart installs

TimWhiting:
In newer installs of dart, pub is now a subcommand of dart rather than it's own executable. 
This causes several melos commands to fail if the user doesn't have the old pub command on their path, and only the new `dart pub` subcommand. I don't know everywhere you are using the old `pub` executable, but I've fixed the places that I have found, and verified it works with at least the `melos bootstrap` command in one of my workspaces.

Included is a new utility function to check if the user has the old `pub` executable available or not.


Ehesp:
Oh nice thanks for this!

Salakar (approved):
Thanks for this, LGTM!﻿

Salakar:
This is available in 0.3.10-dev.3, thanks

PR #27, opened 2020-11-16
feat: Add `--yes` flag to `melos version` for ci support.

cyanglaz:
The version command requires a prompt which blocks it to be run on CI.
This PR adds a `yes` flag so the prompt can be skipped.

This also fixed a format issue.

### DIFF
--- a/packages/melos/lib/src/command/version.dart
+++ b/packages/melos/lib/src/command/version.dart
@@ -62,6 +62,9 @@ class VersionCommand extends Command {
         negatable: true,
         help:
             'By default, melos version will commit changes to pubspec.yaml files and tag the release. Pass --no-git-tag-version to disable the behavior.');
+    argParser.addFlag('yes',
+       defaultsTo: false,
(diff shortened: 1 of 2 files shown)


cyanglaz:
Line 249 has something like 
```
await MelosCommandRunner.instance.run(['run', 'version']);
```

This seems to run the same command twice, is it intentional?

Salakar (approved):
LGTM - thanks for this, will sort CI issue separately and then get this merged & released

Salakar:
> Line 249 has something like
> 
> ```
> await MelosCommandRunner.instance.run(['run', 'version']);
> ```
> 
> This seems to run the same command twice, is it intentional?
> 
> cc @Salakar

This is running the the named script defined in the `melos.yaml` thats named `version`, not the same as `melos version`, it's the equivalent of `melos run version` (it's for running a custom script/step pre/post versioning happening).

cyanglaz:
@Salakar Ah, that makes sense, thanks!

PR #28, opened 2020-11-19
chore: switch to new GitHub action path syntax

Salakar:
Fixes current CI by remove deprecated GitHub action syntax.


Salakar:
Ignoring format failure as fixed in #27

PR #33, opened 2020-11-25
fix(version): re-word the help message of the --yes args in version command.

cyanglaz:
Fix the help message so that the style is more alined with other help messages.


Ehesp (approved):

Salakar (approved):

Salakar:
Thanks

PR #35, opened 2020-11-28
chore(ci): update github actions dependency versions

mikehardy:
I was just in this area for react-native-firebase, so I looked at the actions here and thought I would reverse some entropy for you

checkout was v2 in one place, v1 in others, semantic PR title
was updated - I harmonized it with react-native-firebase version
since that was tested and working post-upgrade

Note:
- I don't have access to invertase/melos.git so couldn't make an in-repo branch, but I don't mind a fork workflow :-)
- the `.github/workflow/scripts` should really be github actions, those are general purpose "install dart", "install flutter" items, I'm surprised they are not? This one looks popular https://github.com/marketplace/actions/flutter-action

### DIFF
--- a/.github/workflows/pr_title.yaml
+++ b/.github/workflows/pr_title.yaml
@@ -1,5 +1,4 @@
-name: pr_title
-
+name: 'PR Title is Conventional'
 on:
   pull_request_target:
     types:
@@ -8,9 +7,9 @@ on:
       - synchronize
 
 jobs:
-  validate:
+  main:
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v1.2.0
+      - uses: amannn/action-semantic-pull-request@v2.1.0
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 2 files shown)


Salakar:
Thanks for sending this up

> * the `.github/workflow/scripts` should really be github actions, those are general purpose "install dart", "install flutter" items, I'm surprised they are not? This one looks popular [github.com/marketplace/actions/flutter-action](https://github.com/marketplace/actions/flutter-action)

We had that action initially on FlutterFire, would hang often or take a long time and also didn't support all flutter channels (e.g. https://github.com/subosito/flutter-action/issues/51) - it was a lot more flakey compared to a simple and quick `git clone --depth 1` of a branch.

Salakar (approved):

mikehardy:
Interesting re the flutter install action not being great. It looks like it does do a lot of feeling around for valid channels etc. Support for master channel is in that action now, so that's solved, but if it's flaky it's flaky.

Speaking of which: even accessing git to github is occasionally flaky in a github runner (I have seen it!) so if the scripts do that, I'd advocate wrapping with nick-invision/retry@v2 also. If you're interested I can PR

PR #45, opened 2020-12-31
feat: add `flutter` package filter

QuirijnGB:
Exposes a `flutter` package filter to the CLI and yaml config.

My specific use case is running tests `flutter test`/`dart test`

### DIFF
--- a/packages/melos/lib/src/common/workspace.dart
+++ b/packages/melos/lib/src/common/workspace.dart
@@ -134,6 +134,8 @@ class MelosWorkspace {
     bool skipPrivate,
     bool published,
     bool hasFlutter,
+    List<String> dependsOn,
+    List<String> noDependsOn,
   }) async {
     if (packages != null) return Future.value(packages);
     final packagePatterns = config.packages;
@@ -248,9 +250,31 @@ class MelosWorkspace {
 
     // --flutter / --no-flutter
     if (hasFlutter != null) {
-      packages = packages
-          .where((package) => package.isFlutterPackage == hasFlutter)
-          .toList();
+      if (hasFlutter) {
+        dependsOn.add("flutter");
+      } else {
+        noDependsOn.add("flutter");
+      }
+    }
+
+    // --depends-on
+    if (dependsOn.isNotEmpty) {
+      packages = packages.where((package) {
+        return dependsOn.every((element) {
+          return package.dependencies.containsKey(element) ||
+              package.devDependencies.containsKey(element);
+        });
+      }).toList();
+    }
+
+    // --no-depends-on
+    if (noDependsOn.isNotEmpty) {
+      packages = packages.where((package) {
+        return noDependsOn.every((element) {
+          return !package.dependencies.containsKey(element) &&
+              !package.devDependencies.containsKey(element);
+        });
+      }).toList();
     }
 
     return packages;
--- a/packages/melos/lib/src/common/workspace.dart
+++ b/packages/melos/lib/src/common/workspace.dart
@@ -133,6 +133,7 @@ class MelosWorkspace {
     List<String> fileExists,
     bool skipPrivate,
     bool published,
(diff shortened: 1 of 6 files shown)


packages/melos/lib/src/common/workspace.dart
```diff
@@ -246,6 +246,18 @@ class MelosWorkspace {
       packagesNoScope = packages;
     }
 
+    // --flutter / --no-flutter
+    if (hasFlutter != null) {
+      var pool = Pool(10);
```
Salakar, line 251: A pool isn't required on this one, guessing you copied pooling logic from the above 'since' flag, that one is a bit more intense and has calls to git cli hence why it's pooled.

But for this one you should be able to do a simpler `where` on the list;

```dart
packages = packages.where((package) => package.isFlutterPackage == hasFlutter).toList();
```

(untested quickly typed code 😅 )

Salakar:
I like it, and I can see the use case. I did originally want to do a more generic `--[no-]depends-on=<depName>` filter instead of this one - should be just as straight forwards (I think - can read `package.dependencies[depName]` && `package.devDependencies[depName]`) - thoughts?

packages/melos/lib/src/common/workspace.dart
QuirijnGB: Even better!
Implemented and tested my usecases :ok_hand:

QuirijnGB:
> I like it, and I can see the use case. I did originally want to do a more generic `--[no-]depends-on=<depName>` filter instead of this one - should be just as straight forwards (I think - can read `package.dependencies[depName]` && `package.devDependencies[depName]`) - thoughts?

Ah yea that makes sense too! 
Thoughts on having both?

Salakar:
> > I like it, and I can see the use case. I did originally want to do a more generic `--[no-]depends-on=<depName>` filter instead of this one - should be just as straight forwards (I think - can read `package.dependencies[depName]` && `package.devDependencies[depName]`) - thoughts?
> 
> Ah yea that makes sense too!
> Thoughts on having both?

No strong objections to both, `--[no-]flutter` could just be sugar around pushing a `flutter` filter onto the  `depends-on` list. If you're willing to contribute both I'd gladly accept it, otherwise we can stick with `flutter` only for now and revise internal code later to use `depends-on`.

Salakar (approved):

Salakar:
Happy to ship as is, lmk if you're happy to ship or if you'd like to look at `--[no-]depends-on` as well

QuirijnGB:
Ill have a look at adding it in

QuirijnGB:
Added a `depends-on` implementation that allows to filter on multiple dependencies. `multiOption` doesnt support `negatable` so had to duplicate a little code. Thoughts?

Salakar:
> Added a `depends-on` implementation that allows to filter on multiple dependencies. `multiOption` doesnt support `negatable` so had to duplicate a little code. Thoughts?

Nice, I like it, goes nicely with the `list` command to quickly find packages with or without certain dependencies

Salakar (approved):
LGTM

Salakar:
Docs preview: https://docs.page/invertase/melos~45/commands

(need to make a preview bot for docs.page sometime 🙈 )

QuirijnGB:
Give it a whirl and if you're happy, I'm happy

Salakar:
Gave it a whirl, works well;

image

Thanks for this :) will ship a release shortly

Salakar:
Published via https://github.com/invertase/melos/commit/a1f45b7e9eaf86cae479e30a471a14a87bec50b2

Thanks again :)

QuirijnGB:
Gotta love a 31st Dec release! (Atleast in Australia :D)

Salakar:
> Gotta love a 31st Dec release! (Atleast in Australia :D)

3:40am here 🙈  but still 31st!

Issue #48, opened 2021-01-05
Feature: melos changed


QuirijnGB:
Are there currently any plans to implement something like [`lerna changed`](https://lerna.js.org/#command-changed)?
It's something that I might need down the line and was wondering if it's something that has been scoped out and would be feasible in Melos.
Use case:
Multiple Flutter apps dependent on common packages. When a common package is changed i would only like to build the affected apps

Salakar:
I can see this being useful. We already have the internal utils in place to allow adding this as we have the --since flag etc so it should  be relatively straightforward to implement I think

If you're willing, a PR will be welcome

QuirijnGB:
Commands:
`melos changed` should default to look back to last tag
`melos changed <tag/commit>`

Fetching the list of affected packages:
1. Find all the affected files up to a given commit/tag
2. Gather all the packages
3. Tie files back to packages
4. List the affected packages

Thoughts?

PR #49, opened 2021-01-07
feat: add `changed` command WIP 

QuirijnGB:
Implements `melos changed <commit/tag>`

Currently does:
* Fetch all the changed files up to the given tag.
* Figure out what packages have been affected
* Trace back the packages that depend on the changed packages
* List the package

TODO:

- [ ] Default behaviour

Resolves: #48

### DIFF
--- a/packages/melos/lib/src/command/changed.dart
+++ b/packages/melos/lib/src/command/changed.dart
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2016-present Invertase Limited & Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this library except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import 'dart:io';
+
+import 'package:ansi_styles/ansi_styles.dart';
+import 'package:args/command_runner.dart' show Command;
+import 'package:melos/src/common/logger.dart';
+import 'package:melos/src/common/package.dart';
+import 'package:melos/src/common/workspace.dart';
+
+class ChangedCommand extends Command {
+  @override
+  final String name = 'changed';
+
+  @override
+  final String description =
+      'Lists all the packages that have been affected by changes since a given commit/tag.';
+
+  @override
+  final String invocation = 'melos changed <commit/tag>';
+
+  @override
+  void run() async {
+    if (argResults.rest.isEmpty) {
+      logger.stderr(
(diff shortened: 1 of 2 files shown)


Salakar:
Out of interest what does this do over `melos list --since=<ref>` - looks like it would produce identical output?

I thought the changed command would be more just `melos changed` or even a new filter for all commands e.g. `melos list --changed` and detected changed the same way the version command does (by finding packages and their dependents that have changed since their last version/tag).

There's also already utils to get commits if you need it; e.g. : https://github.com/invertase/melos/blob/master/packages/melos/lib/src/command/version.dart#L125-L126

The other consideration is should changed be for any change, or specifically packages that have a versionable change, e.g. `fix/feat/etc` vs `chore` commit types; like the version command does; https://github.com/invertase/melos/blob/master/packages/melos/lib/src/command/version.dart#L136-L144

QuirijnGB:
> Out of interest what does this do over `melos list --since=<ref>` - looks like it would produce identical output?

From what I can gather, right now it's similar with the addition that it also outputs all the dependents
 
> I thought the changed command would be more just `melos changed` or even a new filter for all commands e.g. `melos list --changed` and detected changes the same way the version command does (by finding packages and their dependents that have changed since their last version/tag rather than a manual commit sha/ref).

I think `melos changed`'s output could be used as an input for `melos version`. From what I can tell thats what Lerna does as well. For me the addition of the `since` is useful for internal test/nightly builds that don't really get versioned

> There's also already utils to get commits if you need it; e.g. : https://github.com/invertase/melos/blob/master/packages/melos/lib/src/common/git.dart#L138-L141 (note the default behaviour without `since` provided is to detect commits since the last version tag as mentioned previously)

I saw that and decided not to use it because the output was more than I needed and using it would mean that it would run a git command per package. In my case it runs once and outputs all the affected files. I think thats a more scalable approach with many packages, happy to discuss though

> The other consideration is should changed be for any change, or specifically packages that have a versionable change, e.g. `fix/feat/etc` vs `chore` commit types; like the version command does; https://github.com/invertase/melos/blob/master/packages/melos/lib/src/command/version.dart#L136-L144

Yea thats a good idea

Salakar:
> I saw that and decided not to use it because the output was more than I needed and using it would mean that it would run a git command per package. In my case it runs once and outputs all the affected files. I think thats a more scalable approach with many packages, happy to discuss though

Thats fair, I think the original reason I did it per package is so that git sub modules remain supported, e.g. packages/some_package could be a git repository in itself - we have a use case for doing this internally potentially so didn't want to rule it out. I'm ok to ignore our internal use case but there may be projects using melos with submodules in the wild now - so I'm not sure (though it is a new feature and not breaking anything existing) 

> I think melos changed's output could be used as an input for melos version. From what I can tell thats what Lerna does as well. For me the addition of the since is useful for internal test/nightly builds that don't really get versioned

Sure. Long term I'd also like to melos to support how Lerna can do versions with additional args; e.g.:

```
# Repos that use a single version for all packages.
melos version [ custom_version | major | minor | patch | prerelease | nullsafety]
# e.g. `melos version major`

# Bump a specific package (and update all other packages that depend on it to use the new version)
melos version [package_name] [ custom_version | major | minor | patch | prerelease | nullsafety]
# e.g. `melos version firebase_core 0.9.0-nullsafety.0`
```

So that's one to keep in mind also

PR #59, opened 2021-02-15
docs: Add gql-dart/gql as a user of melos

klavs:
`gql-dart/gql` is in the process of adopting melos.

Should be merged after the following PR is finalized
https://github.com/gql-dart/gql/pull/192


Salakar:
Accidentally made ready for review, ignore 🙈

Issue #44, opened 2020-12-26
Feature: allow versioning private packages


ajbarry:
Melos appears to be just what I was looking for to help with the release process for a couple of flutter apps that have a shared (private) package. I have been looking through the docs, examples, issues etc but I am unable to figure out if our current use-case is supported. Here is a melos.yaml config that I am hoping is supported but I can't seem to use the `melos version` command to bump and generate the app changelogs as I hoped:

```
name: MyMonoRepo
packages:
  - app_common/**  # shared common package (publish_to: none)
  - admin_app/**   # flutter admin app, depends on app_common (publish_to: none)
  - user_app/**    # flutter end-user app, depends on app_common (publish_to: none)
environment:
  sdk: ">=2.7.0 <3.0.0"
  flutter: ">=1.24.0-0.0.pre <1.25.0"
```

The result of my command line session yields:

```console
❯ melos list --all --scope="user_app"
user_app

❯ melos version --scope="user_app"
melos version
   └> /Users/me/workspace/mymonorepo

No packages were found that required versioning.
Hint: try running "melos list" with the same filtering options to see a list of packages that were included.
```

I have tried all combinations I can think of using `--all`, `--private`, `--no-published` etc.

Is this a matter of misconfiguration, filesystem layout or something just not supported at this time?

Salakar:
> for a couple of flutter apps that have a shared (private) package

This is definitely supported and a common use case that melos supports as you described. 

However, versioning works for pub.dev only and skips private packages (I imagine your apps have `publish_to: none` on their pubspec.yaml files). Not sure what your use case here for versioning private apps but I guess we could make a change to allow the version bump to still occur for private packages. This is where it gets excluded so you should be able to test it out easily locally by removing these lines: https://github.com/invertase/melos/blob/master/packages/melos/lib/src/command/version.dart#L122-L124 (though we'd probably need to put this behind a flag so as not to be a breaking change).

Everything else apart from versioning should all be working though, with your shared package correctly interlinked across your apps in Dart.

QuirijnGB:
I have a similar use case, multiple customer facing apps that depend on common packages. Ideally you'd be able to bump and version a specific app for release

Salakar:
> I have a similar use case, multiple customer facing apps that depend on common packages. Ideally you'd be able to bump and version a specific app for release

If we can get that line I mentioned above changed to not skip private packages then if you change the common package, and run `melos version` - melos will bump all packages (or in your case apps) that depend on the common package

[...]

PR #67, opened 2021-03-12
feat: allow private packages to be versioned

essarn:
PR for feature request: #44 

Private packages can now be versioned by including the `--all [-a]` flag when running the `version` command.
Also added documentation for this new flag.

### DIFF
--- a/packages/melos/lib/src/command/version.dart
+++ b/packages/melos/lib/src/command/version.dart
@@ -307,7 +307,6 @@ class VersionCommand extends Command {
       return gitCommitsForPackage(
         package,
         since: globalResults['since'] as String,
-        versionAll: versionAll,
       ).then((commits) {
         packageCommits[package.name] = commits
             .map((commit) => ConventionalCommit.parse(commit.message))
@@ -385,7 +384,7 @@ class VersionCommand extends Command {
       logger.stdout(AnsiStyles.gray(
           'Hint: try running "melos list" with the same filtering options to see a list of packages that were included.'));
       logger.stdout(AnsiStyles.gray(
-          'Hint: try running "melos version --all" to include private packages'));
+          'Hint: try running "melos version --all" to include private packages.'));
       return;
     }
 
--- a/packages/melos/lib/src/command/version.dart
+++ b/packages/melos/lib/src/command/version.dart
@@ -91,6 +91,12 @@ class VersionCommand extends Command {
       help: 'Skip the Y/n prompt at the beginning of the command. Applies only '
           'to Conventional Commits based versioning.',
     );
+    argParser.addFlag(
+      'all',
+      abbr: 'a',
+      negatable: false,
+      help: 'Verion private packages that are skipped by default.',
+    );
     argParser.addOption(
       'preid',
       help:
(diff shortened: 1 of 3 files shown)


packages/melos/lib/src/common/git.dart
```diff
@@ -137,9 +137,13 @@ Future<void> gitCommit(String message, {String workingDirectory}) async {
...
-  if (package.isPrivate) {
+Future<List<GitCommit>> gitCommitsForPackage(
+  MelosPackage package, {
+  String since,
+  String preid = 'dev',
+  bool versionAll,
+}) async {
+  if (!versionAll && package.isPrivate) {
     return [];
   }
```
Salakar, line 148: I think it'd be safe to remove the private checking logic here and the new entirely as the only 2 calls that use it already do either private/no-private filters or `all` filters themselves

packages/melos/lib/src/command/version.dart
```diff
@@ -365,15 +373,19 @@ class VersionCommand extends Command {
...
+          .toList();
+    }
 
     if (pendingPackageUpdates.isEmpty) {
       logger.stdout(AnsiStyles.yellow(
           'No packages were found that required versioning.'));
       logger.stdout(AnsiStyles.gray(
           'Hint: try running "melos list" with the same filtering options to see a list of packages that were included.'));
+      logger.stdout(AnsiStyles.gray(
+          'Hint: try running "melos version --all" to include private packages'));
```
Salakar, line 388:
```suggestion
          'Hint: try running "melos version --all" to include private packages.'));
```

Salakar:
Nice, thanks so much for sending this. 2 very minor points but after them this looks good to me

packages/melos/lib/src/common/git.dart
essarn: Yes you are right, removed the check.

Salakar (approved):

essarn:
No problem, happy to help. 
Currently doing a refactor of a project with `melos` and really enjoying the package.

Salakar:
> Currently doing a refactor of a project with melos and really enjoying the package.

Good to hear :D too many late nights went into it so I'm glad it's being used

Salakar:
`0.4.7` has been published with this in, though you now have a local copy setup now which could be good to keep for potential future contributions 🤓

essarn:
> `0.4.7` has been published with this in, though you now have a local copy setup now which could be good to keep for potential future contributions 🤓

Definitely! An interesting thing to do in the future could be to extend the package to version and publish single package projects (not a monorepo) like [cider](https://pub.dev/packages/cider) does.

Salakar:
Oh interesting, not seen cider before, interesting approach. I can see the change log commands being useful or you could also in theory run cider commands themselves on each package via `melos exec -- coder ...`

Also being able to specify a version bump type (as an alternative to specifying exact version), e.g.  `melos version <packageName> <breaking|major|minor|patch|build|packageVersion>` would be cool to support on melos - been on my todo for a while just not got around to it

essarn:
> Oh interesting, not seen cider before, interesting approach. I can see the change log commands being useful or you could also in theory run cider commands themselves on each package via `melos exec -- coder ...`
> 
> Also being able to specify a version bump type (as an alternative to specifying exact version), e.g. `melos version <packageName> <breaking|major|minor|patch|build|packageVersion>` would be cool to support on melos - been on my todo for a while just not got around to it

The version bump feature would for sure be nice to support 👀 

Not a fan of the log command/system though, prefer to just base changelog generation and versioning straight off conventional commits.

Issue #69, opened 2021-03-24
Feature request: Configurable commit messages


shyndman:
Hello there,

This would be a helpful feature for me. Are you open to contributions?

I would likely approach this similarly to Lerna's approach — a configurable `message` field somewhere in melos.yaml, probably associated with the command.

Thanks!

Salakar:
> Are you open to contributions?

Yes, definitely. I left a todo in place for this very thing here: https://github.com/invertase/melos/blob/master/packages/melos/lib/src/command/version.dart#L532-L536

shyndman:
Cool! I'll take a stab at it. :)

shyndman:
Hey @Salakar, can we get this deployed?

Salakar:
On it right now

shyndman:
Thank you :)

Salakar:
Done

shyndman:
Thanks dude.

PR #72, opened 2021-03-25
feat: Support configurable commit messages in version command

shyndman:
The `version` command's `--message` option can be optionally supplied to set a custom commit message when updating versions. If not provided, the previous default is used.

The option's value is treated as a template with a single template var, `{new_package_versions}`, which is replaced with the list of newly versioned packages. Additionally, escaped newlines are evaluated as proper newlines, making it easier to supply a multiline string to the option.

Example
```sh
melos version --message='chore(emergency_release): Hotfix is getting deployed now!\n\n{new_package_versions}'
```

I decided to include a template library to perform the substitution, and picked one that appears to be well-built and small — `mustache_templates`. I could've written something quick and dirty instead, but it didn't feel right to have it live here.

The option is only a first step — we could go further with this, and support defaults in the `melos.yaml` file (like Lerna does with their config), but the options seems like a good first step.

Related to #69

### DIFF
--- a/packages/melos/lib/src/command/version.dart
+++ b/packages/melos/lib/src/command/version.dart
@@ -20,6 +20,7 @@ import 'dart:io';
 import 'package:ansi_styles/ansi_styles.dart';
 import 'package:args/command_runner.dart' show Command;
 import 'package:conventional_commit/conventional_commit.dart';
+import 'package:mustache_template/mustache.dart';
 import 'package:pool/pool.dart' show Pool;
 import 'package:pub_semver/pub_semver.dart';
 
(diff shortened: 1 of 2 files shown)


packages/melos/lib/src/command/version.dart
```diff
@@ -95,7 +115,7 @@ class VersionCommand extends MelosCommand {
       'all',
       abbr: 'a',
       negatable: false,
-      help: 'Verion private packages that are skipped by default.',
+      help: 'Version private packages that are skipped by default.',
```
Salakar, line 118: Woops 👀

packages/melos/lib/src/command/version.dart
```diff
@@ -19,6 +19,7 @@ import 'dart:io';
 
 import 'package:ansi_styles/ansi_styles.dart';
 import 'package:conventional_commit/conventional_commit.dart';
+import 'package:mustache_template/mustache.dart';
```
Salakar, line 22: Seems like a reasonable choice for templating 👍

Salakar (approved):

Salakar:
Thanks for this useful addition 🥳

Salakar:
Tested this out with the latest release of `melos` - for some reason the line breaks didn't come through correctly, might be the wrong format for git, not sure: https://github.com/invertase/melos/commit/928deca384c4b5dc472329163fbef6738fd2426d

shyndman:
Can you bring up the exact command you wrote? I'll get this fixed up.

shyndman:
Nevermind, I repro'd. Give me a few.

shyndman:
I see the issue. I'm not sure how I got this working yesterday, but I was in a bit of shell hell, trying to figure out newlines.

Fix coming right up.

shyndman:
#73

Issue #3, opened 2020-05-25
Document environment variables


Salakar:
Docs need doing for the currently supported environment variables automatically injected by the `exec` command:

### Defined by Melos

- `MELOS_ROOT_PATH`
  - The path to the mono repo root.
- `MELOS_PACKAGE_NAME`
  - The name of the package thats currently executing a script (when using `melos exec`).
- `MELOS_PACKAGE_VERSION`
  - The version of the package thats currently executing a script (when using `melos exec`).
- `MELOS_PACKAGE_PATH`
  - The path of the package thats currently executing a script (when using `melos exec`).
- `MELOS_PARENT_PACKAGE_NAME`
  - The name of the parent package of the current package thats currently executing a script (when using `melos exec`).
- `MELOS_PARENT_PACKAGE_VERSION`
  - The version of the parent package of the current package thats currently executing a script (when using `melos exec`).
- `MELOS_PARENT_PACKAGE_PATH`
  - The path of the parent package of the current package thats currently executing a script (when using `melos exec`).

These variables are available as environment variables in scripts run by `melos exec`. They can also automatically replaced/injected on `melos exec` commands e.g. `melos exec --file-exists="./test_driver/MELOS_PARENT_PACKAGE_NAME_e2e.dart" --scope="*example*" -- echo hello`

#### What is a 'parent package'?

If a package exists in a directory that is also a child of another package, then its parent is the 'parent package'. For example; the package `firebase_auth` has an `example` directory that is also a package, when running a melos exec script in the `example` package then the parent package would be `firebase_auth`.

### User defined

- `MELOS_PACKAGES`
  - Define a comma delimited list of package names that Melos should focus on. This bypasses all filtering flags if defined.

PR #86, opened 2021-04-01
refactor: Move to a stubbable Platform abstraction

shyndman:
This introduces the use of the `platform` package so that we can stub out the `Platform` for tests.

`currentPlatform` should be used in place of direct `Platform` access. It can be overridden through a Zone value, which is useful in testing scenarios (see included test for an example).

**Side note**: This is where Dart's lack of custom lints can be felt. In an ideal system, I'd want to set up a rule where direct use of `Platform` warned the user.

### DIFF
--- a/packages/melos/lib/src/common/utils.dart
+++ b/packages/melos/lib/src/common/utils.dart
@@ -92,26 +92,10 @@ bool get isCI {
       keys.contains('RUN_ID');
 }
 
-String getMelosRoot() {
-  if (currentPlatform.script.path.contains('global_packages')) {
-    return joinAll([
-      File.fromUri(currentPlatform.script).parent.parent.parent.parent.path,
-      'hosted',
-      'pub.dartlang.org',
-      'melos-$melosVersion'
-    ]);
-  }
-
-  // This allows us to use melos on itself during development.
-  if (currentPlatform.script.path.contains('melos_dev.dart')) {
-    return joinAll([
-      File.fromUri(currentPlatform.script).parent.parent.path,
-      'packages',
-      'melos'
-    ]);
-  }
-
-  return File.fromUri(currentPlatform.script).parent.parent.path;
+Future<String> getMelosRoot() async {
+  final melosPackageFileUri = await Isolate.resolvePackageUri(melosPackageUri);
+  // Get from lib/melos.dart to the package root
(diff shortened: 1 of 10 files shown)


Salakar:
#84 has been merged and shipped in `0.4.10-dev.1`.

----


This looks like a very useful addition to making things testable. 🎉

packages/melos/test/package_test.dart
```diff
@@ -0,0 +1,69 @@
...
+
+      final package = workspace.packages.first;
+      await package.getPublishedVersions();
+
+      expect(interceptor.isDone, isTrue);
+    });
+
+    test(
+      'requests published packages from PUB_HOSTED_URL if present',
+      withMockPlatform(
```
Salakar, line 53: Nice - glad this is tested

Salakar (approved):
I know it's a draft but LGTM - being able to test envs is going to be super helpful especially for all these; https://github.com/invertase/melos/issues/3

packages/melos/lib/src/common/utils.dart
```diff
@@ -82,20 +85,33 @@ bool promptBool({String message = 'Continue?', bool defaultsTo = false}) {
...
+  // This allows us to use melos on itself during development.
+  if (currentPlatform.script.path.contains('melos_dev.dart')) {
+    return joinAll([
+      File.fromUri(currentPlatform.script).parent.parent.path,
+      'packages',
+      'melos'
+    ]);
+  }
+
+  return File.fromUri(currentPlatform.script).parent.parent.path;
```
Salakar, line 114:
```suggestion
Future<String> getMelosRoot() async {
  final melosPackageFileUri = await Isolate.resolvePackageUri(melosPackageUri);
  // Get from lib/melos.dart to the package root
  return File(melosPackageFileUri.toFilePath()).parent.parent.path;
```

packages/melos/lib/src/common/utils.dart
```diff
@@ -26,6 +26,9 @@ import 'package:prompts/prompts.dart' as prompts;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:yaml/yaml.dart';
 
+import '../../version.g.dart';
```
Salakar, line 29:
```suggestion
```

PR #123, opened 2021-06-04
fix: exec hang, exec trailing options

### LINKED ISSUE #15: melos exec hang

2ZeroSix:
closes #15
closes #16
inspired by #16

- pass script source directly to shell using `-c` option
- ignore trailing options in exec command (exec tried to parse all available options including those passed to the child script)
  same approach is used in `pub run` https://github.com/dart-lang/pub/blob/291705ca0b9632cb945dd39493dd5b9db41b897a/lib/src/command/run.dart#L27

### DIFF
--- a/packages/melos/lib/src/common/utils.dart
+++ b/packages/melos/lib/src/common/utils.dart
@@ -215,7 +215,9 @@ Future<int> startProcess(
 
   final execProcess = await Process.start(
     executable,
-    currentPlatform.isWindows ? ['/C', '%MELOS_SCRIPT%'] : [],
+    currentPlatform.isWindows
+        ? ['/C', '%MELOS_SCRIPT%']
+        : ['-c', filteredArgs.join(' ')],
     workingDirectory: workingDirectoryPath,
     environment: {
       ...environment,
@@ -224,13 +226,6 @@ Future<int> startProcess(
     },
     runInShell: true,
   );
-
-  if (!currentPlatform.isWindows) {
-    // Pipe in the arguments to trigger the script to run.
-    execProcess.stdin.writeln(filteredArgs.join(' '));
-    // Exit the process with the same exit code as the previous command.
-    execProcess.stdin.writeln(r'exit $?');
-  }
 
   var stdoutStream = execProcess.stdout;
   var stderrStream = execProcess.stderr;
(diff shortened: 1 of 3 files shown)


Salakar (approved):

packages/melos/lib/src/common/utils.dart
```diff
@@ -213,25 +213,19 @@ Future<int> startProcess(
     return _arg;
   }).where((element) => element != null);
 
+  final scriptSource = filteredArgs.join(' ');
   final execProcess = await Process.start(
     executable,
-    currentPlatform.isWindows ? ['/C', '%MELOS_SCRIPT%'] : [],
+    currentPlatform.isWindows ? ['/C', '%MELOS_SCRIPT%'] : ['-c', scriptSource],
```
Salakar, line 219:
```suggestion
    currentPlatform.isWindows ? ['/C', '%MELOS_SCRIPT%'] : ['-c', r'$MELOS_SCRIPT'],
```

Wouldn't this work? Keeps it consistent with how Windows does it too then

2ZeroSix: It causes wrong variable expansion, you can see in test:
`bin/sh: 1: "dart: not found`

single argument `"dart analyze . --fatal-infos"`
becomes `"dart` `analyze` `.` `--fatal-infos"`

2ZeroSix: @Salakar ^

Salakar: I see now, will revert, one moment

Salakar: What's your thoughts on `['-c', r'eval "$MELOS_SCRIPT"']` - since this does work

Salakar: Have tested on Melos and FlutterFire and everything seems to be working with the above so will merge - I can change in another PR if needed. Thanks for your help :)

PR #127, opened 2021-06-08
fix: enable Dart SDK for root IntelliJ project

### LINKED ISSUE #124: Question: Pre-defined configurations in IntelIj are removed when running melos bootstrap
Hey @Salakar 
I have pre-defined configurations in my IntelIj IDE like running a different flavor of my app(dev, staging, prod). But when I run the `melos bs` command. Those configurations are removed.
Screenshot 2021-06-08 at 1 49 13 PM

I only see these are added back but I lose my own configurations.

SAGARSURI:
When you run `melos bootstrap` command it doesn't enable Dart SDK for the root module. This fixes the issue.

Fixes #124

### DIFF
--- a/packages/melos/templates/intellij/modules/workspace_root_module.iml.tmpl
+++ b/packages/melos/templates/intellij/modules/workspace_root_module.iml.tmpl
@@ -6,5 +6,7 @@
       <sourceFolder url="file://$MODULE_DIR$" isTestSource="false" />
     </content>
     <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" name="Dart SDK" level="project" />
+    <orderEntry type="library" name="Dart Packages" level="project" />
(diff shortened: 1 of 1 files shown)


SAGARSURI:
I just have one question though. Since my root module was a Flutter plugin. I also saw this line getting added to the `melos_*.yaml` file
```
<orderEntry type="library" name="Flutter Plugins" level="project" />
```

Do you think this is also required? I tried without adding this and it still worked.

Salakar:
> I just have one question though. Since my root module was a Flutter plugin. I also saw this line getting added to the `melos_*.yaml` file
> 
> ```
> <orderEntry type="library" name="Flutter Plugins" level="project" />
> ```
> 
> Do you think this is also required? I tried without adding this and it still worked.

I don't think it's needed

PR #155, opened 2021-08-06
feat: add support for printing current melos version via `-v` or `--version`

Salakar:


packages/melos/lib/src/command_runner.dart
```diff
@@ -55,4 +63,14 @@ class MelosCommandRunner extends CommandRunner<void> {
     addCommand(PublishCommand());
     addCommand(VersionCommand());
   }
+
+  @override
+  Future<void> runCommand(ArgResults topLevelResults) async {
+    if (topLevelResults['version'] == true) {
+      // ignore: avoid_print
+      print(melosVersion);
```
rrousselGit, line 71: This probably should use the logger

packages/melos/lib/src/command_runner.dart
```diff
@@ -55,4 +63,14 @@ class MelosCommandRunner extends CommandRunner<void> {
     addCommand(PublishCommand());
     addCommand(VersionCommand());
   }
+
+  @override
+  Future<void> runCommand(ArgResults topLevelResults) async {
```
Salakar, line 68: Overriding since this is the recommended way to do global flag in command runner: 

image

packages/melos/lib/src/command_runner.dart
Salakar: logger is a instance property of `MelosCommand` whereas this isn't a command instance 🤔

Salakar: Leaving as is for now as discussed on slack

Issue #154, opened 2021-08-03
The plugin `X` doesn't have a main class


btrautmann:
Hi there! Our team has been doing a spike on what it would look like to migrate our project to using `melos`. We've started by moving all our packages into `packages` which we specify within the `melos.yaml`. So far, everything has worked. When we finally moved the last package which is our actual `app` package and `main` executable, we started getting errors along the lines of:

```
The plugin `example` doesn't have a main class defined in /path/to/example/ExamplePlugin.java or /path/to/example/ExamplePlugin.kt. This is likely to due to an incorrect `androidPackage: com.example` or `mainClass` entry in the plugin's pubspec.yaml.
If you are the author of this plugin, fix the `androidPackage` entry or move the main class to any of locations used above. Otherwise, please contact the author of this plugin and consider using a different plugin in the meanwhile. 

• app
    └> ./packages/app
    └> Failed to run "flutter pub get" in this package.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Bootstrap failed: "flutter pub get" failed in one of your workspace packages.
```

It's worth noting that `flutter pub get` in the actual package that's being called out here succeeds.

While trying to debug this, I removed the plugin in question as a dependency from all packages that were declaring it as one, and the error moved on to _another_ local plugin, so I'm thinking it's not specific to the plugin that originally had the error.

We're currently using Flutter `2.2.2` on the stable channel and Dart `2.13.3`. 

Thank you in advance for any help!

Salakar:
Hey are you able to provide an example of what the pubspec looks like for your app package?

btrautmann:
> Hey are you able to provide an example of what the pubspec looks like for your app package?

I created a gist [here](https://gist.github.com/btrautmann/770137f77b1f8a2b6598e31ff81e6f51). I redacted any specific names, but left the structure un-touched!

btrautmann:
@Salakar any thoughts or avenues you'd recommend taking to figure this out?

btrautmann:
[...]

PR #162, opened 2021-08-27
fix(bootstrap): Allow add-to-app packages to bootstrap

btrautmann:
When trying to integrate Melos into our project, we were running into an issue running `melos bootstrap`. That issue is detailed in https://github.com/invertase/melos/issues/154.

TLDR: Bootstrapping fails and we receive the below error, but running `flutter pub get` directly in the failing package succeeds:

```
The plugin `example` doesn't have a main class defined in /path/to/example/ExamplePlugin.java or /path/to/example/ExamplePlugin.kt. This is likely to due to an incorrect `androidPackage: com.example` or `mainClass` entry in the plugin's pubspec.yaml.
If you are the author of this plugin, fix the `androidPackage` entry or move the main class to any of locations used above. Otherwise, please contact the author of this plugin and consider using a different plugin in the meanwhile.
```

This error can be found in the flutter tooling [here](https://github.com/flutter/flutter/blob/c99ed373b3d254f9b10789ddf3771c3aa39e48ca/packages/flutter_tools/lib/src/platform_plugins.dart#L148-L156).

During `boostrap`, melos creates a mirror directory in `<workspace_path>/.dart_tool/.melos_tool`, and copies pubspec information to this directory in the same path structure as the project itself. However, the assertion above requires the main class defined in each plugin's `pubspec.yaml` to also be there, which is what this PR achieves.

A couple notes:

1. This only seems to be an issue for Android. I'll look at the code linked above in the flutter tooling to confirm this, but anecdotally, all our plugins also support iOS and we don't have any issues with the iOS side of things.

2. I haven't written any kind of testing yet, because I wanted to vet this approach with the maintainers prior to going that far. The most testing I've done is ensuring that `melos bootstrap` now succeeds for us, where as before it was failing.


packages/melos/lib/src/commands/bootstrap.dart
```diff
@@ -219,6 +219,40 @@ Future<void> _generateTemporaryProjects(MelosWorkspace workspace) async {
...
+          // at the same relative location, so pub get can find it
+          final hasJavaPluginClass = plugin.javaPluginClassPath != null;
+          final hasKotlinPluginClass = plugin.kotlinPluginClassPath != null;
+          final pathParts = plugin.androidPackage!.split('.');
+          final mainClassDirectoryName = hasJavaPluginClass ? 'java' : 'kotlin';
+          final mainClassFileSuffix = hasJavaPluginClass ? '.java' : '.kt';
+          final destinationMainClassPath = joinAll([
+            join(workspace.melosToolPath, plugin.pathRelativeToWorkspace),
+            'android/src/main/$mainClassDirectoryName',
+            pathParts.join('/'),
```
lesnitsky, line 243:
```suggestion
            ...pathParts,
```

packages/melos/lib/src/commands/bootstrap.dart
```diff
@@ -219,6 +219,40 @@ Future<void> _generateTemporaryProjects(MelosWorkspace workspace) async {
...
+          ]);
+          File(destinationMainClassPath).createSync(recursive: true);
+          if (hasJavaPluginClass) {
+            File(plugin.javaPluginClassPath!)
+                .copySync(destinationMainClassPath);
+          }
+          if (hasKotlinPluginClass) {
+            File(plugin.kotlinPluginClassPath!)
+                .copySync(destinationMainClassPath);
+          }
```
lesnitsky, line 254:
```suggestion
          String? classPath;
          if (hasJavaPluginClass) {
              classPath = plugin.javaPluginClassPath!;
          } else if (hasKotlinPluginClass) {
              classPath = plugin. kotlinPluginClassPath!;
          }
          File(classPath!).copySync(destinationMainClassPath);
```

packages/melos/lib/src/package.dart
```diff
@@ -784,6 +791,58 @@ class Package {
...
+    return null;
+  }
+
+  String? get kotlinPluginClassPath {
+    if (androidPackage == null || androidPluginClass == null) return null;
+
+    final kotlinPluginClassPath = joinAll([
+      path,
+      'android/src/main/kotlin',
+      androidPackage!.split('.').join('/'),
```
lesnitsky, line 838:
```suggestion
      ...androidPackage!.split('.'),
```

lesnitsky (changes requested):

btrautmann:
@lesnitsky @rrousselGit @Salakar I implemented the changes requested by @lesnitsky as well as added logic to the existing bootstrap test that adds an add-to-app module and android-supporting flutter plugin to the workspace under test. The test passes, and when writing it I confirmed that it did fail if I commented out my new logic, with the following error:

```
BootstrapException: failed to install c at /Users/brandontrautmann/Development/projects/melos-fork/packages/melos/kZ8rcZ/packages/c.
package:melos/src/commands/bootstrap.dart 154:9  _BootstrapMixin._runPubGet
===== asynchronous gap ===========================
package:melos/src/commands/bootstrap.dart 31:11  _BootstrapMixin.bootstrap.<fn>
===== asynchronous gap ===========================
package:melos/src/commands/runner.dart 117:7     _Melos._runLifecycle
===== asynchronous gap ===========================
test/commands/bootstrap_test.dart 43:7           main.<fn>.<fn>
```

Please let me know if the test additions aren't what you want. Due to the nature of creating the projects individually via `createProject`, I opted for this approach of reaching into the `unParsedYaml` to figure out if we should create a main class. If desired, I can create a more isolated test that doesn't use `createProject` and instead explicitly creates just the add to app module and flutter plugin.

lesnitsky (approved):
LGTM code-wise, can't tell anything about the approach since I'm a bit out of the context

btrautmann:
> LGTM code-wise, can't tell anything about the approach since I'm a bit out of the context

Thanks for the review! Would it make sense for one of the other reviewers to give the 👍 before merging?

Salakar (approved):

Issue #163, opened 2021-08-27
Proposal: match unknown commands with scripts


jpnurmi:
I've often noticed myself accidentally trying to run commands like `melos analyze` or `melos test`, or whatever scripts there are in various projects. It would be convenient if `melos <script>` would run the script in case its name doesn't conflict with any built-in commands.

PR #140, opened 2021-06-20
feat: added config validation and type-safe Dart API (#139)

rrousselGit:

PR #167, opened 2021-08-31
feat: Match unknown commands with scripts

jpnurmi:
Allow running `melos <script>` as a shortcut for `melos run <script>`, provided that the script name doesn't conflict with built-in commands.

```yaml
name: test_workspace
packages: [packages/**]
scripts:
  test: echo "running test script..."
  clean: echo "running clean script..."
```

In the above workspace, `melos test` is equivalent to `melos run test`:
```
$ melos test
melos run test
   └> echo "running test script..."
       └> RUNNING

running test script...

melos run test
   └> echo "running test script..."
       └> SUCCESS
```

However, `melos clean` is a built-in command:
```
$ melos clean
Cleaning workspace...

Workspace cleaned. You will need to run the bootstrap command again to use this workspace.
```

Close: #163

### DIFF
--- a/packages/melos/lib/src/command_runner.dart
+++ b/packages/melos/lib/src/command_runner.dart
@@ -24,6 +24,7 @@ import 'command_runner/exec.dart';
 import 'command_runner/list.dart';
 import 'command_runner/publish.dart';
 import 'command_runner/run.dart';
+import 'command_runner/script.dart';
 import 'command_runner/version.dart';
 import 'common/utils.dart';
 import 'workspace_configs.dart';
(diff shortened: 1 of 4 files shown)


jpnurmi:
The problem is that https://github.com/invertase/melos/pull/140 postponed workspace config reading until commands get run. However, in order to be able to inject scripts as commands, the workspace config needs to be read before command-line arguments are parsed. As such, this ends up reading the workspace config twice. There probably should be a way to somehow pass the pre-parsed workspace config to individual commands.

rrousselGit:
> The problem is that #140 postponed reading the workspace config until commands get run.

We could add a constructor on `Melos` to instantiate it from an existing configuration.

Then it's a matter of doing `addCommand(ScriptCommand(configs))`

Salakar:
Are you able to add a test by any chance? We should also probably switch one of the CI jobs to use this new short form, e.g. `analyze`: https://github.com/invertase/melos/blob/master/.github/workflows/validate.yaml#L18

+1 on the suggestion by @rrousselGit above - are you able to make this change also? 

Thanks!

jpnurmi:
TODO: tests :slightly_smiling_face:

Salakar (approved):

PR #169, opened 2021-09-04
refactor: Instantiate workspace from configs

jpnurmi:
This prevents the workspace config from being parsed multiple times under certain scenarios. At the same time, it helps to implement a cleaner solution to match unknown commands with scripts (#163).

### DIFF
--- a/packages/melos/lib/src/workspace.dart
+++ b/packages/melos/lib/src/workspace.dart
@@ -48,15 +48,12 @@ class MelosWorkspace {
     required this.logger,
   });
 
-  /// Build a [MelosWorkspace] from a Directory.
-  /// If the directory is not a valid Melos workspace (e.g. no "melos.yaml" file)
-  /// then null is returned.
-  static Future<MelosWorkspace> fromDirectory(
-    Directory directory, {
+  /// Build a [MelosWorkspace] from a workspace configuration.
+  static Future<MelosWorkspace> fromConfig(
+    MelosWorkspaceConfig workspaceConfig, {
     PackageFilter? filter,
     required Logger logger,
   }) async {
-    final workspaceConfig = await MelosWorkspaceConfig.fromDirectory(directory);
     final allPackages = await PackageMap.resolvePackages(
       workspacePath: workspaceConfig.path,
       packages: workspaceConfig.packages,
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -74,8 +74,8 @@ mixin _RunMixin on _Melos {
     };
 
     if (script.filter != null) {
-      final workspace = await MelosWorkspace.fromDirectory(
-        Directory(config.path),
+      final workspace = await MelosWorkspace.fromConfig(
+        config,
         filter: script.filter,
         logger: logger,
       );
--- a/packages/melos/lib/src/commands/runner.dart
+++ b/packages/melos/lib/src/commands/runner.dart
@@ -85,8 +85,9 @@ abstract class _Melos {
       );
     }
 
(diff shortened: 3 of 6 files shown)


packages/melos/lib/src/workspace.dart
```diff
@@ -48,15 +48,12 @@ class MelosWorkspace {
     required this.logger,
   });
 
-  /// Build a [MelosWorkspace] from a Directory.
-  /// If the directory is not a valid Melos workspace (e.g. no "melos.yaml" file)
-  /// then null is returned.
-  static Future<MelosWorkspace> fromDirectory(
```
rrousselGit, line 54: I would keep both constructors. I don't think there's a real value in changing existing usage of `fromDirectory` to `fromConfig`.

We could instead of `fromDirectory` use `fromConfig`

jpnurmi: For #163, the config has to be read early on before creating a workspace or even before parsing command-line arguments. Even though `MelosWorkspace.fromDirectory` is a convenient way to construct a workspace, calling it would result in parsing the config again.

rrousselGit: Yes I agree. But #163 is an edge-case. For more common cases, `fromDirectory` is fine, and switching to `fromConfig` is unnecessary

jpnurmi: The next step would be to read the config once at the top and pass it down to the command runner: https://github.com/jpnurmi/melos/commit/c98f9fbc8e76ee262c791a67bfed5b35d0d99d38

Since a workspace config provides the workspace path, there's no need to assume `Directory.current` down the tree anymore. Does this make it sound more reasonable to always instantiate workspaces from that config?

The last step (#163) would then look something like this: https://github.com/jpnurmi/melos/commit/8c6a5a55671d61f1fdfb514d8b73670628e12c86

jpnurmi: By the way, if you, for any reason, would prefer to solve #163 yourself, I'm happy to hand it over to you. :)

rrousselGit: Thanks for the reply!

> Since a workspace config provides the workspace path, there's no need to assume Directory.current down the tree anymore.

I'm not sure that this is a good thing. 
I could see commands have an argument for changing the working directory / specifying the path to melos.yaml, in which case we'd have to revert the change.

Similarly, MelosWorkspace.fromDirectory is a public API. Removing it is a breaking change, as people can call Melos commands programatically instead of relying on the CLI

jpnurmi: > I could see commands have an argument for changing the working directory / specifying the path to melos.yaml, in which case we'd have to revert the change.

Sounds even better reason to keep track of the top-level config path.

> Similarly, MelosWorkspace.fromDirectory is a public API. Removing it is a breaking change, as people can call Melos commands programatically instead of relying on the CLI

Aren't you about to release a new major version?

Salakar: > > Similarly, MelosWorkspace.fromDirectory is a public API. Removing it is a breaking change, as people can call Melos commands programatically instead of relying on the CLI
> 
> Aren't you about to release a new major version?

This API isn't public yet on a non-dev release yet since we're introducing it as part of v1 - so I don't think this would be a breaking change yet 🤔

rrousselGit: Oh, my bad I thought we already had a stable 1.0 release. You can ignore me then.

rrousselGit (approved):
LGTM

Salakar (approved):

PR #176, opened 2021-09-17
refactor: Pass workspace config from the top

jpnurmi:
To be able to match unknown commands with scripts (#163), the list of scripts must be known before parsing command-line arguments.

Thus, instead of reading the workspace config in individual commands, read the config early on startup and pass it down to the command runner and commands so that:

  a) commands don't need to read the config again, and
  b) it will be straightforward to introduce a new command that matches unknown commands with scripts.


Salakar (approved):
LGTM - thanks again

PR #178, opened 2021-09-17
feat: Added an error message when multiple packages in the workspace have the same name

### LINKED ISSUE #174: `PackageMap.resolvePackages` does not add all the `example` packages in the map.
Hey, 
First of all, thanks for creating `melos`. We are using it in [StreamChatFlutter](https://github.com/GetStream/stream-chat-flutter) and it makes our life easier 😄 

We recently tried to use the latest [pre-release](https://pub.dev/packages/melos/versions/1.0.0-dev.3) version of `melos` and we saw some differences in the `boostrap` command. Previously it used to successfully bootstraps all our `packages` and their `example` apps. But now, it bootstraps all our packages but not all our package `example` apps.

So I looked into the [implementation](https://github.com/invertase/melos/blob/master/packages/melos/lib/src/package.dart#L320-L340) and found that rather than bootstrapping all the packages and examples. It now only bootstraps unique packages.

eg:
```yaml
packages:
	- stream_chat
	- stream_chat/example
	- stream_chat_flutter
	- stream_chat_flutter/example
```
We expect that `melos` will bootstrap all the provided packages. But it only bootstraps `3` packages, overwriting the `example` because of the same name.
```dart
final name = pubSpec.name!;

packageMap[name] = Package._(...);
```

rrousselGit:
closes #174

### DIFF
--- a/packages/melos/lib/src/package.dart
+++ b/packages/melos/lib/src/package.dart
@@ -23,6 +23,7 @@ import 'package:cli_util/cli_logging.dart';
 import 'package:collection/collection.dart';
 import 'package:glob/glob.dart';
 import 'package:http/http.dart' as http;
+import 'package:melos/src/common/validation.dart';
 import 'package:path/path.dart';
 import 'package:pool/pool.dart';
 import 'package:pub_semver/pub_semver.dart';
@@ -323,6 +324,19 @@ class PackageMap {
         final pubSpec = await PubSpec.load(pubspecFile.parent);
 
         final name = pubSpec.name!;
+
+        if (packageMap.containsKey(name)) {
+          throw MelosConfigException(
+            '''
+Multiple packages with the name `$name` found in the workspace, which is unsupported.
+To fix this problem, consider renaming your packages to have a unique name.
+
+The packages that caused problem are:
+- $name at ${relative(pubspecDirPath, from: workspacePath)}
+- $name at ${relative(packageMap[name]!.path, from: workspacePath)}
+''',
+          );
+        }
 
         packageMap[name] = Package._(
           name: name,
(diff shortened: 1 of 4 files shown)


packages/melos/lib/src/package.dart
```diff
@@ -324,6 +325,19 @@ class PackageMap {
 
         final name = pubSpec.name!;
 
+        if (packageMap.containsKey(name)) {
+          throw MelosConfigException(
+            '''
+Multiple packages with the name `$name` found in the workspace, which is unsupported.
+To fix this problem, consider renaming your packages to have a unique name.
+
+The packages that caused problem are:
```
Salakar, line 334:
```suggestion
The packages that caused the problem are:
```

packages/melos/test/workspace_test.dart
```diff
@@ -29,6 +30,46 @@ import 'utils.dart';
...
+        () async => MelosWorkspace.fromConfig(
+          await MelosWorkspaceConfig.fromDirectory(workspaceDir),
+          logger: TestLogger(),
+        ),
+        throwsMelosConfigException(
+          message: '''
+Multiple packages with the name `example` found in the workspace, which is unsupported.
+To fix this problem, consider renaming your packages to have a unique name.
+
+The packages that caused problem are:
```
Salakar, line 65:
```suggestion
The packages that caused the problem are:
```

packages/melos/test/workspace_test.dart
```diff
@@ -29,6 +30,46 @@ import 'utils.dart';
...
+          logger: TestLogger(),
+        ),
+        throwsMelosConfigException(
+          message: '''
+Multiple packages with the name `example` found in the workspace, which is unsupported.
+To fix this problem, consider renaming your packages to have a unique name.
+
+The packages that caused the problem are:
+- example at packages/b/example
+- example at packages/a/example
```
Salakar, line 67:
```suggestion
- example at packages/a/example
- example at packages/b/example
```

PR #192, opened 2021-11-05
fix: melos.yaml ignores should apply also to `run` commands `MELOS_PACKAGES` env variable

Salakar:
Fixes FlutterFire CI erroring trying to run tests for packages that are actually ignored in melos.yaml

### DIFF
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -129,18 +129,6 @@ mixin _RunMixin on _Melos {
       // a defined script, this comma delimited list of package names is used
       // instead of any filters if detected.
       environment[envKeyMelosPackages] = packagesEnv;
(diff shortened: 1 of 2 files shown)


packages/melos/lib/src/commands/run.dart
```diff
@@ -126,6 +129,18 @@ mixin _RunMixin on _Melos {
...
+        config,
+        filter: PackageFilter(ignore: config.ignore),
+        logger: logger,
+      );
+      final packages = workspace.filteredPackages.values.toList();
+      // MELOS_PACKAGES environment is detected by melos itself when through
+      // a defined script, this comma delimited list of package names is used
+      // instead of any filters if detected.
+      environment[envKeyMelosPackages] =
+          packages.map((e) => e.name).toList().join(',');
```
Salakar, line 143:
```suggestion
```

PR #193, opened 2021-11-08
fix: run version cmd with `--depentens-versions` value from cli

blaugold:
Currently the value specified for `--depentens-versions` on the command line is not passed through to the `autoVersion` method.


Salakar (approved):
Good spot, thanks for the PR 🙂

PR #194, opened 2021-11-09
feat: respect exact version constraints when updating dependents

blaugold:
If a dependent of a package declares an exact version constraint for that dependency, updating that version constraint as part of versioning, now replaces the old constraint with an exact constraint for the new version.


blaugold:
Thanks for merging. Probably not a common use case, but it helps me :)

Salakar:
> Thanks for merging. Probably not a common use case, but it helps me :)

No problem, am loving all the contributions :)

Issue #149, opened 2021-07-17
melos publish should publish packages in dependency order


rrousselGit:
In some cases, `melos publish` can publish packages in an order different than the dependency order.

For example, a monorepo may have:

- A, a package with no dependency
- B, which depends on A
- C, which also depends on A

In this case, A should be published first. Then B/C can be published in no special order

This ensures that pub.dev analysis doesn't fail because it trying to analyze the B before A was published

PR #199, opened 2021-12-01
feat: Add topological sort to publish command

dnys1:
Fixes https://github.com/invertase/melos/issues/149

Adds topological sorting to the `publish` command so that packages may be published in order based on their dependencies on other workspace packages.


Ehesp:
Thanks for this. The team is pretty busy right now but we'll get this reviewed ASAP.

Salakar:
Thanks for this, I've taken a look already and it looks great. My only concern is having a `CycleException` being thrown is a breaking changing, some repositories have cyclic dependencies, e.g. FlutterFire - so this will break publishing. Is there any way to sort what we can and any that can't be sorted due to cyclic dependencies just bump them to the top/bottom of the ordering?

dnys1:
Thanks, @Salakar! I think the best we can do in that case is to leave packages with inter-dependencies untouched, since they cannot be topologically sorted. I've updated the code to, in essence, ignore cyclic errors. Let me know how that sounds.

Anything more complicated such as sorting a subgraph of packages which do not have inter-dependencies is possible - although I'm not sure how worthwhile the investment would be. Let me know!

Salakar:
> Thanks, @Salakar! I think the best we can do in that case is to leave packages with inter-dependencies untouched, since they cannot be topologically sorted. I've updated the code to, in essence, ignore cyclic errors. Let me know how that sounds.
> 
> Anything more complicated such as sorting a subgraph of packages which do not have inter-dependencies is possible - although I'm not sure how worthwhile the investment would be. Let me know!

This is perfect for now, thanks for addressing review points

dnys1:
Awesome - thanks @Salakar!

Issue #43, opened 2020-12-15
Documentation: building a local app pointing to local packages repo


ffeu:
Hi

I couldn't find in the docs the right way to build a local app pointing to the local packages of a melos monorepo. 

Using `flutterfire` as an example, say I have an app named `app1` with this `pubspec.yaml`:

`app1/pubspec.yaml`
```yaml
...
dependencies:
  flutter:
    sdk: flutter
  firebase_core: 0.5.3
  firebase_auth: 0.18.4+1
```

But I've cloned `flutterfire` locally and I want to test some stuff in the `app1` with the local code, so without `melos` I would do something like this:

```yaml
...
dependencies:
  flutter:
    sdk: flutter
  firebase_core: 
    path: "../flutterfire/packages/firebase_core/firebase_core"
  firebase_auth: 
    path: "../flutterfire/packages/firebase_auth/firebase_auth" 
```

But that doesn't work. I tried bootstraping (`melos bootstrap`) in one, both repos (`flutterfire` and `app1`), but couldn't make it work. 

Actually, is the manual path override necessary at all in `app1`?

I'm confident there must be a way, but I just couldn't find it in the docs and didn't manage to do it myself.

Thanks in advance

Ehesp:
Within your `melos.yaml`, could you try adding the pages to the glob lookup? Need to check whether it supports out of directory paths though:

```
name: Melos

packages:
  - packages/**
  - "../flutterfire/packages/**"

```

Salakar:
Melos works from packages nested from the root of the mono repo unfortunately. So you won't be able to add a packages glob going outside of the directory

However you can still create a relative symlink to a directory outside of your mono repo root which should work as melos does support symlinks (and it can also be committed and would look similar to [this](https://github.com/FirebaseExtended/flutterfire/blob/master/packages/firebase_core/firebase_core/macos/Classes/FLTFirebaseCorePlugin.m));


```bash
git clone https://github.com/FirebaseExtended/flutterfire.git
cd mono_repo_dir
# relative symlink - which can be commited to git
ln -s ../flutterfire flutterfire
```

Update melos.yaml:

```yaml
name: Melos

packages:
  - packages/**
  - flutterfire/**
```

You may also wish to add some more explicity packages definitions for only the packages you're using - for less overhead;


```yaml
name: Melos

packages:
  - packages/**
  - flutterfire/firebase_auth/**
  - flutterfire/firebase_core/**
# etc ...
```

Salakar:
[...]

PR #211, opened 2021-12-11
feat: follow symlinks when resolving packages

Quijx:
This PR makes it so that projects are found if they are reachable through paths containing links that are allowed in the `melos.yaml`.

This change is roughly implemented as discussed here: https://github.com/invertase/melos/issues/43#issuecomment-746203744

A few notes:
- I have added a check that the found `FileSystemEntity` is a `File`. That should stop problems with folders named `pubspec.yaml` (Even if likely nobody ever would do this. :smile:)
- I used the `basename` function from the `path` package instead of `file.path.endsWith('pubspec.yaml')` to check if the file is named `pubspec.yaml` to avoid problems with files called for example `old_pubspec.yaml`
- I could not use `.toSet()` as you suggested in https://github.com/invertase/melos/issues/43#issuecomment-746203744 because `File` does not override `==` such that they are equal if they point to the same file.
- If a `pubspec.yaml` is reachable through one or more links and directly, either path could be added to `allPubspecs`. I assumed it is not important which one is taken. Is that assumption correct, or should i add a check to use the direct path if it is available?
- I have not thought through any implications this change might have because projects outside of the project root are now allowed. I am not familiar enough with the inner workings of melos, so I'll let you think that through :smile:


Quijx:
So I have another idea about how to find the files matching the globs in `packages`: Instead of listing the contents of the workspace directory and filtering the ones matching the globs, one could use the `listFileSystem` function on the globs themselves. That could be faster in some cases because only paths need to be followed that even have the chance of matching the glob. However some paths might be visited multiple times if the globs overlap. This would also have the implication that paths outside the root can be added as packages, for example with the glob `../outside_package/*`. Is that something you would want? Should be easy to implement.

Quijx:
So I fixed the lints that caused the `analyze` job to fail, but I'm not so sure why the `test_macos` job (https://github.com/invertase/melos/runs/4505402142?check_suite_focus=true) failed. Locally the test `test/commands/bootstrap_test.dart: bootstrap resolves workspace packages with path dependency` passes for me and sadly I don't have a macos computer to test it. I could use some help here...

Salakar (approved):

Salakar:
I think this is ok to go in it's current form, unless you have any reason it should not be merged yet?

I also want to test it out on FlutterFire manually to make sure all is well

Quijx:
I think it's fine in its current form. If you think there will be performance problems because some people might link large parts of the file system that have to be searched on every bootstrap, I can look into the `glob.listFileSystem` solution. But i expect most people will only link the project folders, so it's probably fine. So you can merge if you want.

Salakar (approved):

PR #221, opened 2022-01-05
docs: add Flame to projects using Melos

spydon:
The order was somewhat by GitHub stars I think, is it okay to have Flame second or should I add it last? :)


Ehesp (approved):

spydon:
Oh, ofc it should be `docs:`, still new to  conventional commits.

PR #242, opened 2022-02-06
feat: add `--version` option to `version` command

### LINKED ISSUE #229: No entries generated for changelog
When we ([Flame](https://github.com/flame-engine/flame)) run for example:
```sh
melos version flame_tiled 1.1.1
```
there is no automatic changelog entries generated, even though there are unreleased conventional commit in the git log.
(No difference if it is run with `-c`)

See the commit log here:
https://github.com/flame-engine/flame/commits/main
`#1324` should for example get a changelog line generated for it right?

Most likely us that are doing something wrong, but would be great with some guidance. :slightly_smiling_face:

blaugold:
Closes #229

## Description

This change adds an option to the `version` command which allows the user to manually specify a version for a package. The option can be used multiple times.

The user is asked to provided a changelog message for packages for which a version has been manually specified. If there are no commits to describe the changes for such a package, the changelog message is mandatory. Otherwise it is optional.

Using the `--version` option for a package versions it regardless of whether it was included by the package filters or whether there are commits since the last version.

With this change the alternative invocation (`melos version <package> <version>`) becomes redundant and could be removed.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/common/workspace_changelog.dart
+++ b/packages/melos/lib/src/common/workspace_changelog.dart
@@ -58,12 +58,14 @@ class WorkspaceChangelog {
         .where((update) => update.reason == PackageUpdateReason.graduate);
     final packagesWithBreakingChanges = pendingPackageUpdates.where(
       (update) =>
-          update.reason == PackageUpdateReason.commit &&
+          (update.reason == PackageUpdateReason.commit ||
+              update.reason == PackageUpdateReason.manual) &&
           update.semverReleaseType == SemverReleaseType.major,
     );
     final packagesWithOtherChanges = pendingPackageUpdates.where(
       (update) =>
-          update.reason == PackageUpdateReason.commit &&
+          (update.reason == PackageUpdateReason.commit ||
+              update.reason == PackageUpdateReason.manual) &&
           update.semverReleaseType != SemverReleaseType.major,
     );
 
--- a/docs/commands/version.mdx
+++ b/docs/commands/version.mdx
@@ -7,7 +7,7 @@ description: "Learn more about the `version` command in Melos."
 
 > Supports all [Melos filtering](/filters) flags.
 
-Automatically version and generate changelogs for all packages. 
+Automatically version and generate changelogs for all packages.
 
 
 ```bash
@@ -107,4 +107,13 @@ chore(release): publish packages
 
 ```bash
 melos version --message="chore(release): publish new versions"
+```
+
+## --version (-V)
+
+Manually specify a version for a package. Can be used multiple times. Each value must be in the format "package:version". Applies only to Conventional Commits based versioning. Cannot be combined with --graduate or --prerelease flag.
+
+```bash
+melos version --version=foo:1.0.0
+melos version -V=foo:1.0.0
 ```
--- a/docs/guides/automated-releases.mdx
+++ b/docs/guides/automated-releases.mdx
@@ -29,6 +29,7 @@ TODO docs on:
 - build versions
 - dependency versioning
 - bad commit messages?
+- manual versioning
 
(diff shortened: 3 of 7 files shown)


Salakar:
Amazing, thanks for this. Will get to reviewing.

> With this change the alternative invocation (melos version <package> <version>) becomes redundant and could be removed.

I'm not sure this is the case, `melos version <package> <version>` versions only the package mentioned in this command, whereas it seems `--version` adds the package mentioned to other packages that need versioning (unless I'm misunderstanding how the new `--version` flag works)

packages/melos/lib/src/common/workspace_changelog.dart
```diff
@@ -58,12 +58,14 @@ class WorkspaceChangelog {
         .where((update) => update.reason == PackageUpdateReason.graduate);
     final packagesWithBreakingChanges = pendingPackageUpdates.where(
       (update) =>
-          update.reason == PackageUpdateReason.commit &&
+          (update.reason == PackageUpdateReason.commit ||
+              update.reason == PackageUpdateReason.manual) &&
           update.semverReleaseType == SemverReleaseType.major,
```
Salakar, line 63: Based on https://github.com/invertase/melos/pull/242/files#diff-8a6755f20b64cdbfb7a9290c81f41a53b6b9d63fb4f7c60e1fa4641cce172572R145-R152 - manual updates would never be `SemverReleaseType.major` so is Is this change needed?

Salakar:
I've not tested but I'm wondering if there's any conflict with the global `--version` flag; https://github.com/invertase/melos/blob/main/packages/melos/lib/src/command_runner.dart#L53-L82

blaugold:
> I'm not sure this is the case, melos version <package> <version> versions only the package mentioned in this command, whereas it seems --version adds the package mentioned to other packages that need versioning (unless I'm misunderstanding how the new --version flag works)

Yeah, you're right. 
But I think it should now be possible to implement `melos version <package> <version>` in terms of `_VersionMixin.autoVersion` with the `manualVersions` argument.

> I've not tested but I'm wondering if there's any conflict with the global --version flag;

I didn't think about that global flag :) Probably best to rename `--version` to `--manual-version`.

packages/melos/lib/src/common/workspace_changelog.dart
blaugold: Manually versioned packages can have commits, but if they don't, we still need a `SemverReleaseType`. That's what that check is for. If a manually versioned package has commits, some of them can be breaking and `MelosPendingPackageUpdate .semverReleaseType` would return `SemverReleaseType.major`.

Maybe a better approach would be to make `MelosPendingPackageUpdate.semverReleaseType` nullabe and return `null` for manually versioned packages. If `semverReleaseType` is `null` we determine whether to include the breaking change note in the changelog based on if there are any breaking change commits.

Salakar:
> I've updated a few things to address your comments.

Amazing thank you. I'll sanity test this on FlutterFire's release today just to check normal versioning doesn't have any issues, after that I think we can ship this 🎉

blaugold:
Just FYI, this is breaking the programmatic Melos API because in `_VersionMixin` `version` was removed and `versionAuto` was renamed to `version`.

blaugold:
Good idea! We can also handle `build` version increments. Otherwise, I would limit that feature to packages with non-prerelease versions

Salakar:
Yep, all the logic for versioning/next versioning is already in place and has lots of tests too (+ follows Dart conventions + community conventions about shifting versioning down for pre 1.0.0 versions)

Also found a small bug with manual versioning, the root/workspace change log for manual versioned packages show no changes; 

image

Issue #159, opened 2021-08-20
Update git tag of dependant packages


danilofuchs:
[_setDependencyVersionForPackage](https://github.com/invertase/melos/blob/master/packages/melos/lib/src/commands/version.dart#L394) currently ignores git dependencies entirely.

It would be nice to check if a git dependency is targeting an internal package and update its git tag constraint as well. We currently have to do this manually, parsing the output of `melos version` and manually editing the tags before applying the versioning

```diff
dependencies:
  internal_dep:
    git:
      url: git@github.com:org/repo.git
      path: packages/internal_dep
-    ref: internal_dep-v0.0.1
+    ref: internal_dep-v0.0.2
```

fstof:
Indeed this would be handy

fstof:
Hope the PR suffices...

"Works on my computer" 😅

blaugold:
With #256 being merged, can we close this issue?

PR #256, opened 2022-03-03
feat: support git hosted package dependencies when versioning

fstof:
## Description
This change implements the feature request #159 

When packages are hosted on git (ie, not published). When we call `melos version` the package's `pubspec.yaml` git ref tags should be updated to the latest git tag created.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/docs/guides/automated-releases.mdx
+++ b/docs/guides/automated-releases.mdx
@@ -54,7 +54,7 @@ Once satisfied with your pending releases, release them to [pub.dev](https://pub
 melos publish --no-dry-run
 ```
 
-## Git hosted packages
+## git hosted packages
 
 To use this feature enable it in `melos.yaml`
 ```yaml
--- a/docs/guides/automated-releases.mdx
+++ b/docs/guides/automated-releases.mdx
@@ -54,7 +54,7 @@ Once satisfied with your pending releases, release them to [pub.dev](https://pub
 melos publish --no-dry-run
 ```
 
-## git hosted packages
+## Git hosted packages
 
 To use this feature enable it in `melos.yaml`
 ```yaml
@@ -74,7 +74,7 @@ dependencies:
       path: packages/internal_dep
       ref: internal_dep-v0.0.1
 
-# to be updated to
+# will be updated to:
 
 dependencies:
   internal_dep:
--- a/docs/guides/automated-releases.mdx
+++ b/docs/guides/automated-releases.mdx
@@ -53,3 +53,26 @@ Once satisfied with your pending releases, release them to [pub.dev](https://pub
 ```dart
 melos publish --no-dry-run
 ```
+
+## Git hosted packages
+
+If your packages are private and don't publish to [pub.dev](https://pub.dev), you can use the tag generated as git reference in your `pubspec.yaml` file and Melos will ensure the versions are updated accordingly.
+
+Example:
+```yaml
+dependencies:
+  internal_dep:
+    git:
+      url: git@github.com:org/repo.git
+      path: packages/internal_dep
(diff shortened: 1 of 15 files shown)


packages/melos/lib/src/commands/version.dart
```diff
@@ -439,10 +441,23 @@ Hint: try running "melos version --all" to include private packages.
...
-        dependencyVersionReplaceRegex(dependencyName), (Match match) {
-      return '${match.group(1)}$dependencyVersion';
-    });
+    String updatedContents;
+
+    final gitReference =
+        package.pubSpec.dependencies[dependencyName] is GitReference ||
+            package.pubSpec.devDependencies[dependencyName] is GitReference;
+
+    if (gitReference) {
```
Salakar, line 450: We should probably only do this if `--git-tag-version` is true right?

Salakar:
Thanks for this, I think if we want to support this we should put it behind an melos.yaml option for the version command, e.g. like these: https://github.com/invertase/melos/blob/main/melos.yaml#L7-L8 - since it's very specific behaviour (and also would break current behaviour). Thoughts?

packages/melos/lib/src/commands/version.dart
fstof: That makes sense yes.
I will check how we do that and update

fstof: Okay @Salakar 
I double checked this flag. (The documentation needs updating as it is just a copy/paste from `--no-changelog`) 

However. It seems like this option still updates the versions in `pubspec.yaml` only skips the git commit (which obviously also does not tag).

If this is indeed the intended behaviour (skipping the commit) then I think we should still do the update, but do introduce the cofig in `melos.yaml`

fstof:
> Thanks for this, I think if we want to support this we should have it turned off by default and put it behind an `melos.yaml` option for the version command, e.g. like these: https://github.com/invertase/melos/blob/main/melos.yaml#L7-L8 - since it's very specific behaviour (and also would break current behaviour). Thoughts?

Hi @Salakar
So I implemented both suggestions, (the `melos.yaml` config and `--git-tag-version` check) you can check if you hare happy with that.

But I am of the opinion that `--git-tag-version` is not required as i mentioned in https://github.com/invertase/melos/pull/256#discussion_r820081584

docs/guides/automated-releases.mdx
```diff
@@ -53,3 +53,33 @@ Once satisfied with your pending releases, release them to [pub.dev](https://pub
 ```dart
 melos publish --no-dry-run
 ```
+
+## git hosted packages
```
Salakar, line 57:
```suggestion
## Git hosted packages
```

docs/guides/automated-releases.mdx
```diff
@@ -53,3 +53,33 @@ Once satisfied with your pending releases, release them to [pub.dev](https://pub
...
+```yaml
+dependencies:
+  internal_dep:
+    git:
+      url: git@github.com:org/repo.git
+      path: packages/internal_dep
+      ref: internal_dep-v0.0.1
+
+# to be updated to
+
```
Salakar, line 78:
```suggestion

# will be updated to:

```

fstof:
> > But I am of the opinion that --git-tag-version is not required as i mentioned in [#256 (comment)](https://github.com/invertase/melos/pull/256#discussion_r820081584)
> 
> Let's remove this check then since it's now going to be behind melos.yaml option anyway

removed the `--git-tag-version` check

fstof:
@Salakar 

Is there anything specific still holding this back?

Salakar (approved):

Issue #117, opened 2021-05-29
The pubspec.yaml file has changed since the pubspec.lock file was generated


rhalff:
I'm trying to run the build runner for code generation, the first package succeeds but the other packages fail to run because the pubspec.yaml has changed since the pubspec.lock was generated. Both of those packages depend on the first one (not sure if that matters).

Bootstrap:
```bash
➜  flutter_json_schema_form git:(master) ✗ melos bootstrap           
melos bootstrap
   └> /home/rhalff/development/flutter_json_schema_form

Running "flutter pub get" in workspace packages...
  ✓ fjsf_theme_material
     └> ./packages/fjsf-theme-material
  ✓ fjsf
     └> ./packages/fjsf
  ✓ flutter_json_schema_form_example
     └> ./packages/example

Linking workspace packages...
  > SUCCESS

Generating IntelliJ IDE files...
  > SUCCESS

 -> 3 plugins bootstrapped
```

Running build_runner in each package directory:
```
➜  flutter_json_schema_form git:(master) ✗ melos run generate:flutter
melos run generate:flutter
   └> melos exec -c 1 --depends-on="build_runner" --flutter -- "flutter pub run build_runner build --delete-conflicting-outputs"
       └> RUNNING

$ melos exec
   └> flutter pub run build_runner build --delete-conflicting-outputs
       └> RUNNING (in 3 packages)

-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
fjsf:
[INFO] Generating build script...
[INFO] Generating build script completed, took 319ms

[INFO] Initializing inputs
[INFO] Reading cached asset graph...
[INFO] Reading cached asset graph completed, took 74ms

[INFO] Checking for updates since last build...
[INFO] Checking for updates since last build completed, took 558ms

[INFO] Running build...
[INFO] Running build completed, took 19ms

[INFO] Caching finalized dependency graph...
[INFO] Caching finalized dependency graph completed, took 53ms

[INFO] Succeeded after 80ms with 0 outputs (0 actions)

fjsf: SUCCESS

-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
fjsf_theme_material:
The pubspec.yaml file has changed since the pubspec.lock file was generated, please run "pub get" again.
pub finished with exit code 65

-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #284, opened 2022-04-27
feat: add support for Dart/Flutter SDK at custom `sdkPath`

### LINKED ISSUE #81: Support custom flutter sdk / dart sdk paths
I'm using [`flutter_wrapper`](https://github.com/passsy/flutter_wrapper) in my projects and it's not possible to pass in a custom flutter sdk path. This causes `bootstrap` to fail because my project isn't compatible with the flutter version in `$PATH`.

This request is similar to #21 but more general as @orestesgaolin wrote https://github.com/invertase/melos/issues/21#issuecomment-745978584

> Another idea is to allow to pass arbitrary path to Flutter installation. This way it wouldn't be tied to fvm only. 

### Suggestion

```bash
melos bootstrap --flutter-sdk=.flutter
```

acoutts:
## Description

This PR adds a new `melos.yaml` configuration setting `sdk_path` which allows someone to specify the path to their dart/flutter sdk instead of relying on what's in the path.
This makes melos compatible with flutter version manager.

Closes #81

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/commands/bootstrap.dart
+++ b/packages/melos/lib/src/commands/bootstrap.dart
@@ -68,24 +68,27 @@ mixin _BootstrapMixin on _CleanMixin {
       );
     }
 
-    for (final package in workspace.allPackages.values) {
-      if (package.pubSpec.dependencyOverrides.isNotEmpty) {
-        logger?.stderr(
-          '''
+    await Stream.fromIterable(workspace.allPackages.values).parallel(
+      (package) async {
+        if (package.pubSpec.dependencyOverrides.isNotEmpty) {
+          logger?.stderr(
+            '''
   $_warningLabel: ${AnsiStyles.bold(package.name)}
     └> dependency_overrides in pubspec.yaml are being overridden in pubspec_overrides.yaml''',
-        );
-      }
+          );
+        }
 
-      await _generatePubspecOverrides(workspace, package);
-      await _runPubGetForPackage(workspace, package);
+        await _generatePubspecOverrides(workspace, package);
+        await _runPubGetForPackage(workspace, package);
 
-      logger?.stdout(
-        '''
+        logger?.stdout(
+          '''
   $_checkLabel ${AnsiStyles.bold(package.name)}
     └> ${AnsiStyles.blue(printablePath(package.pathRelativeToWorkspace))}''',
-      );
-    }
+        );
+      },
+      parallelism: canRunPubGetConcurrently ? null : 1,
+    ).drain<void>();
   }
 
   Future<void> _generatePubspecOverrides(
@@ -150,16 +153,18 @@ mixin _BootstrapMixin on _CleanMixin {
   }
 
   // Return a stream of package that completed.
-  Stream<Package> _runPubGet(MelosWorkspace workspace) async* {
-    for (final package in workspace.filteredPackages.values) {
-      await _runPubGetForPackage(
-        workspace,
-        package,
-        inTemporaryProject: true,
+  Stream<Package> _runPubGet(MelosWorkspace workspace) =>
+      Stream.fromIterable(workspace.filteredPackages.values).parallel(
+        (package) async {
+          await _runPubGetForPackage(
+            workspace,
+            package,
+            inTemporaryProject: true,
+          );
+          return package;
+        },
+        parallelism: canRunPubGetConcurrently ? null : 1,
       );
-      yield package;
-    }
-  }
 
   Future<void> _runPubGetForPackage(
     MelosWorkspace workspace,
--- a/packages/melos/lib/src/common/utils.dart
+++ b/packages/melos/lib/src/common/utils.dart
@@ -115,6 +115,9 @@ String get nextDartMajorVersion {
 bool get isPubspecOverridesSupported =>
     currentDartVersion.compareTo(Version.parse('2.17.0-266.0.dev')) >= 0;
 
+bool get canRunPubGetConcurrently =>
+    currentDartVersion.compareTo(Version.parse('2.16.0')) >= 0;
+
 String promptInput(String message, {String? defaultsTo}) {
   return prompts.get(message, defaultsTo: defaultsTo);
 }
@@ -434,5 +437,48 @@ extension DirectoryUtils on Directory {
     }
 
     return recurse(this, {});
+  }
+}
+
+extension StreamUtils<T> on Stream<T> {
+  /// Runs [convert] for each event in this stream and emits the result, while
+  /// ensuring that no more events than specified by [parallelism] are being
+  /// processed at any given time.
+  ///
+  /// If [parallelism] is `null`, [Platform.numberOfProcessors] is used.
+  Stream<R> parallel<R>(
+    Future<R> Function(T) convert, {
+    int? parallelism,
+  }) async* {
+    final pending = <Future<R>>[];
+    final done = <Future<R>>[];
+
+    await for (final value in this) {
+      late final Future<R> future;
+      future = Future(() async {
+        try {
+          return await convert(value);
+        } finally {
+          pending.remove(future);
+          done.add(future);
+        }
+      });
+      pending.add(future);
+
+      if (pending.length < (parallelism ?? Platform.numberOfProcessors)) {
+        continue;
+      }
+
+      await Future.any(pending);
+
+      for (final future in done) {
+        yield await future;
+      }
+      done.clear();
+    }
+
+    for (final result in await Future.wait(pending)) {
+      yield result;
+    }
   }
 }
--- a/packages/melos/test/commands/bootstrap_test.dart
+++ b/packages/melos/test/commands/bootstrap_test.dart
@@ -83,7 +83,7 @@ void main() {
 
       expect(
         logger.output,
-        equalsIgnoringAnsii(
+        ignoringAnsii(
           '''
 melos bootstrap
    └> ${workspaceDir.path}
@@ -176,21 +176,31 @@ Generating IntelliJ IDE files...
 
       expect(
         logger.output,
-        equalsIgnoringAnsii(
-          '''
+        ignoringAnsii(
+          allOf(
+            [
+              '''
 melos bootstrap
    └> ${workspaceDir.path}
 
-Running "flutter pub get" in workspace packages...
+Running "flutter pub get" in workspace packages...''',
+              '''
   ✓ a
     └> packages/a
+''',
+              '''
   ✓ b
     └> packages/b
+''',
+              '''
   ✓ c
     └> packages/c
+''',
+              '''
   ✓ d
     └> packages/d
-
+''',
+              '''
 Linking workspace packages...
   > SUCCESS
 
@@ -199,6 +209,8 @@ Generating IntelliJ IDE files...
 
  -> 4 plugins bootstrapped
 ''',
+            ].map(contains).toList(),
+          ),
         ),
       );
 
@@ -444,7 +456,7 @@ dependency_overrides:
 
       expect(
         logger.output,
-        equalsIgnoringAnsii(
+        ignoringAnsii(
           '''
 melos bootstrap
    └> ${workspaceDir.path}
--- a/packages/melos/test/commands/exec_test.dart
+++ b/packages/melos/test/commands/exec_test.dart
@@ -51,7 +51,7 @@ void main() {
 
         expect(
           logger.output,
-          equalsIgnoringAnsii(
+          ignoringAnsii(
             '''
 \$ melos exec
    └> echo hello world
--- a/packages/melos/test/commands/list_test.dart
+++ b/packages/melos/test/commands/list_test.dart
@@ -35,7 +35,7 @@ void main() {
 
           expect(
             logger.output,
-            equalsIgnoringAnsii(
+            ignoringAnsii(
               '''
 a
 b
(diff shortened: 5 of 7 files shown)


mikehardy:
Just a question - shouldn't a version manager reconfigure the PATH so that invoked scripts are already the fvm-correct invoked scripts?

I ask because version managers come and go, there always ends up being more than one, and in the react-native ecosystem at least, after a while there were too many of them to support, and they finally scrapped supporting any of them and said "use your version manager before invoking our scripts, if the version manager is doing it's job on PATH correctly, our scripts will work with your version-manager's binaries..."

acoutts:
I see your point, it's not great to make melos opinionated like that.
But the point of fvm is not to change your path at all and to work locally within each project.
If you have 3 flutter projects on your computer, all 3 can have their own flutter version independent of the one in your path- the idea being when you run `fvm flutter ...` in any of those projects it will do so with the correct flutter version that project is intended to use.

It appears that just trying to wrap the melos command with `fvm flutter pub global run melos` doesn't work because melos will simply run `flutter ...` from path which is not the one we want to happen.

acoutts:
Perhaps a better approach is like what https://github.com/invertase/melos/issues/81 is asking for - instead of `use-fvm` it could be an optional path to the flutter executable which would make this compatible with any version manager.

mikehardy:
> But the point of fvm is not to change your path at all and to work locally within each project.
> If you have 3 flutter projects on your computer, all 3 can have their own flutter version independent of the one in your path

:thinking: I see what you mean. It quickly becomes a rabbit-hole, in the node ecosystem it's typically to have 'nvm' to alter PATH, and 'auto-nvm' installed in combination with an .nvmrc or .node_version file in a directory and if auto-nvm sees it it will `nvm use <version from config file>`.

Not sure there even is a good way to do it - it's kind of all hacks :laughing: - I think opening up path config vs having an opinion on a specific tool is better though, personally.

I'm not deeply invested in this though - I just wanted to bring in other-ecosystem experience where other-ecosystem finally just threw up their hands and said "we're not supporting specific version managers" because of proliferation. If I can predict the future here, it's that

acoutts:
Ok i've updated it to just pass an optional `sdkPath` setting either in your melos.yaml or to the bootstrap command.

This way if you use fvm you can just set this in your melos.yaml:
```yaml
sdk_path: .fvm/flutter_sdk
```

packages/melos/lib/src/command_runner/base.dart
```diff
@@ -123,6 +123,12 @@ abstract class MelosCommand extends Command<void> {
           'matches the other filters. The included packages skip --ignore '
           'and --since checks.',
     );
+
+    argParser.addOption(
+      filterOptionSdkPath,
```
blaugold, line 128: I don't think this option should be a filter option. We could add a new `GlobalOptions` object for which the `argParser` is setup in `setupGlobalOptionsParser` and is parsed in `parseGlobalOptions`.

packages/melos/lib/src/commands/bootstrap.dart
```diff
@@ -171,10 +171,29 @@ mixin _BootstrapMixin on _CleanMixin {
     Package package, {
     bool inTemporaryProject = false,
   }) async {
-    final pubGetArgs = ['pub', 'get'];
-    final execArgs = package.isFlutterPackage
-        ? ['flutter', ...pubGetArgs]
-        : [if (utils.isPubSubcommand()) 'dart', ...pubGetArgs];
+    late final List<String> execArgs;
```
blaugold, line 174: Why not just this?
```
final execArgs = [
    if (package.isFlutterApp)
        '${workspace.sdkPath}flutter'
    else if (utils.isPubSubcommand(sdkPath: workspace.sdkPath))
        '${workspace.sdkPath}dart',
    'pub',
    'get',
]
```

Of if `workspac.sdkPath` is nullable:
```
final execArgs = [
    if (package.isFlutterApp)
        '${workspace.sdkPath ?? ''}flutter'
    else if (utils.isPubSubcommand(sdkPath: workspace.sdkPath))
        '${workspace.sdkPath ?? ''}dart',
    'pub',
    'get',
]
```

packages/melos/lib/src/common/utils.dart
```diff
@@ -46,6 +46,7 @@ const filterOptionDependsOn = 'depends-on';
 const filterOptionNoDependsOn = 'no-depends-on';
 const filterOptionIncludeDependents = 'include-dependents';
 const filterOptionIncludeDependencies = 'include-dependencies';
+const filterOptionSdkPath = 'sdk-path';
```
blaugold, line 49: `filterOptionSdkPath` -> `globalOptionSdkPath`

packages/melos/lib/src/workspace.dart
```diff
@@ -98,6 +128,9 @@ class MelosWorkspace {
   late final bool isFlutterWorkspace =
       allPackages.values.any((package) => package.isFlutterPackage);
 
+  /// Optionally specify sdk path for dart/flutter
+  late final String sdkPath;
```
blaugold, line 132: This field does not need to be `late`. 
I would prefere it if this field is nullable to signify that the sdk path is not overriden and not the empty string.

packages/melos/lib/src/workspace.dart
```diff
@@ -137,17 +170,23 @@ class MelosWorkspace {
...
 
   /// Builds a dependency graph of dependencies and their dependents in this workspace.
-  Future<Map<String, Set<String>>> getDependencyGraph() async {
+  Future<Map<String, Set<String>>> getDependencyGraph({
+    required String sdkPath,
+  }) async {
     final pubDepsExecArgs = ['--style=list', '--dev'];
     final pubListCommandOutput = await Process.run(
       isFlutterWorkspace
           ? 'flutter'
```
blaugold, line 179: We should also use the sdkPath here and and wherever else we invoke `flutter` and `dart` to be consistent.

packages/melos/lib/src/workspace_configs.dart
```diff
@@ -418,10 +419,16 @@ class MelosWorkspaceConfig {
       map: yaml,
     );
 
+    final sdkPath = assertKeyIsA<String?>(
+      key: 'sdk_path',
```
blaugold, line 423: `sdk_path` -> `sdkPath`

blaugold (commented):
Currently, we assume that melos runs with the same version of Dart as the `dart` tool that is invoked by Melos:

https://github.com/invertase/melos/blob/1fdff169f6fccda26cc4ea71e4d199adfcf30b54/packages/melos/lib/src/common/utils.dart#L107-L109

`currentDartVersion` needs to become a function that takes the sdk path, invokes `dart --version` and parses its output.

packages/melos/lib/src/command_runner/base.dart
acoutts: Thanks for clarifying. What is a filter option supposed to be used for?

blaugold: The filter options are used to select a subset of the packages in a workspace, for example to only bootstrap Flutter packages and not pure Dart packages you could use `melos bootstrap --flutter`.

acoutts:
I'm experiencing an issue and I can't tell if it's my branch or the v2 release.
There is some kind of race condition because i have to run bootstrap twice on a fresh repo. The first one fails with an error like:
```
Could not find a file named "pubspec.yaml" in "/builds/M5hecXxM/3/bottle/pay/app/.pub-cache/git/mockito-aa841a3c3762afab2c31486677015ff0a9abc2cf".
BootstrapException: Failed to install.: common at /builds/M5hecXxM/3/bottle/pay/app/packages/common.
```
But if I run it again it works.
If I use melos 1.5.x (before 2.x), there are no issues either.

blaugold:
We changed to bootstrapping packages in parallel in 2.0.0, which was disabled because of a race condition in `pub`, which should have been fixed in Dart [`2.16.0`](https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#2160---2022-02-03).

blaugold:
It might be that there is another race condition specific to git dependencies.

acoutts:
Interesting- good context to have. In my case i'm on dart 2.17 from the flutter master channel.
Is there a way to toggle off parallel pub gets?

blaugold:
> Interesting- good context to have. In my case i'm on dart 2.17 from the flutter master channel. Is there a way to toggle off parallel pub gets?

Not as a feature. To validate your code in this PR you could temporarily hard-code to return `false` from `canRunPubGetConcurrently`:
https://github.com/invertase/melos/blob/1fdff169f6fccda26cc4ea71e4d199adfcf30b54/packages/melos/lib/src/common/utils.dart#L118

> In my project, 4 of my packages rely on the same external git dependency in that error message (same version)- maybe it's something with shared git dependencies when run in parallel.

Yeah, that might be it. Once the repo directory exists in the pub cache, `pub` might assume that the package is available. Could open an issue in [`pub`](https://github.com/dart-lang/pub) repo for this?

acoutts:
>Not as a feature. To validate your code in this PR you could temporarily hard-code to return false from canRunPubGetConcurrently:

I can confirm this worked. I filed https://github.com/dart-lang/pub/issues/3404 for it.
I temporarily committed it here just so I can get my build pushed tonight, and tomorrow I will add a flag for it as this will probably be broken for other people as well, and will a future dart sdk fix.

packages/melos/lib/src/common/utils.dart
acoutts: Renamed in 7a9a9151fb73d5a2ee70692d55bcdd73c8c868f4

packages/melos/lib/src/commands/bootstrap.dart
acoutts: Simplified in ca2d3a8a9ed5fec764c3667fad0702c818593a1e (includes tests)

packages/melos/lib/src/workspace.dart
acoutts: Fixed in 79c997306f1e48af542dee1c395274ee51528330, 9364e2ed807faa6eb36779f81c3628efb0b71808 and 17b102a8a496292b1c0dfdf499494859f6f3b0d3

* Added a new function to simplify the logic here assembling the final sdk path
* Using the path package now to correctly append the paths together (to work on all platforms)
* Added tests

packages/melos/lib/src/workspace.dart
acoutts: e94de99c841d914fc34b77ff24b056448925d896

acoutts:
Just pushed some refactoring from your feedback- thanks for that.
The only one i'm not 💯  on how to do is how to convert what i created as a filter setting into a global setting.

packages/melos/lib/src/workspace_configs.dart
acoutts: Fixed in ebe5de74daa428593bc131295a89ca342a7ebcf1

blaugold:
Thanks for this PR, BTW! I have created a separate PR for the option to disable running `pub get` in parallel (#285). I can take care of refactoring to a global setting, if you don't mind.

acoutts:
Cheers! That would be great. Want me to add you as a contributor to my fork or what's easiest for you?

blaugold:
If you have "Allow edits from maintainers" on this PR enabled, I should be able to update the feature branch.

acoutts:
I dont see that option for some reason, but I just invited you as a maintainer to the fork in case that works.

acoutts:
Is there a trick to get around this problem? Every time I run bootstrap, i can't use the flutter pub command:
```
The pubspec.yaml file has changed since the pubspec.lock file was generated, please run "flutter pub get" again.
pub finished with exit code 65
```
I can do `melos exec flutter pub get` but of course that reverts the bootstrap magic.

blaugold:
I think you're running into https://github.com/invertase/melos/issues/117. Until this is fixed, a workaround is to run `melos exec flutter pub get` to let flutter generate temporary files and then bootstrap with melos.

acoutts:
Weird i guess bootstrap never worked for me then! i was using `melos exec flutter pub get` always but learned while doing this PR that i was using it wrong before.

This is what my lockfiles look like for modules when doing the melos pub get:
```
  nisaba:
    dependency: transitive
    description:
      path: "packages/nisaba"
      ref: "nisaba-v1.0.0"
      resolved-ref: "1bcf666152369d89810f2ff340c536131809fae6"
      url: redacted
    source: git
    version: "1.0.0"
```

And then if i run bootstrap the lock file changes to this:
```
  nisaba:
    dependency: "direct overridden"
    description:
      path: "../nisaba"
      relative: true
    source: path
    version: "1.0.0"
```

acoutts:
I guess my workaround for now is to just use `path: ../nisaba` in the pubspec instead of trying to use the git tags system from melos.
It's ok for now as we don't have any external consumers of our packages so paths works. Still sorting out a good app release flow with melos (currently use tags to initiate CI releases but melos would interfere with that).

blaugold:
I just tried to push to `bottlepay/melos/main` but I must have made some kind of mistake. 😅  I'm trying to fix it right now, though.

blaugold:
@acoutts OK, the tip of `bottlepay/melos/main` now has the correct changes. Could open a new PR from that branch with the same title as this one?

acoutts:
Opened https://github.com/invertase/melos/pull/288

PR #288, opened 2022-04-29
feat: add support for Dart/Flutter SDK at custom `sdkPath`

### LINKED ISSUE #81: Support custom flutter sdk / dart sdk paths

acoutts:
## Description

This PR adds support for specifying the path to a Dart/Flutter SDK, instead of relying on what's in the path. 
This makes melos compatible with flutter version manager (fvm).

The SDK path can be specified though the following mechanisms, in order of highest to lowest precedence:

1. `--sdk-path` command line option
2. `MELOS_SDK_PATH` environment variable
3. `sdkPath` in `melos.yaml`

Closes #81

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -31,6 +31,16 @@ Supported hosts:
 repository: https://github.com/invertase/melos
 ```
 
+## `sdkPath`
+
+Path to the Dart/Flutter SDK that should be used, unless overridden though the command line.
+
+Relative paths are resolved relative to the `melos.yaml` file.
+
+```yaml
+sdkPath: .fvm/flutter_sdk
+```
+
 ## `packages`
 
 > required
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -37,6 +37,8 @@ Path to the Dart/Flutter SDK that should be used.
 
 Relative paths are resolved relative to the `melos.yaml` file.
 
+To use the system-wide SDK, provide the special value "auto".
+
 If the SDK path is specified though multiple mechanisms, the precedence from highest to lowest is:
 
 1. `--sdk-path` global command line option
--- a/docs/configuration/overview.mdx
+++ b/docs/configuration/overview.mdx
@@ -33,9 +33,15 @@ repository: https://github.com/invertase/melos
 
 ## `sdkPath`
 
-Path to the Dart/Flutter SDK that should be used, unless overridden though the command line.
+Path to the Dart/Flutter SDK that should be used.
 
 Relative paths are resolved relative to the `melos.yaml` file.
+
+If the SDK path is specified though multiple mechanisms, the precedence from highest to lowest is:
+
+1. `--sdk-path` global command line option
+2. `MELOS_SDK_PATH` environment variable
+3. `sdkPath` in `melos.yaml`
 
 ```yaml
 sdkPath: .fvm/flutter_sdk
--- a/docs/environment-variables.mdx
+++ b/docs/environment-variables.mdx
@@ -48,3 +48,8 @@ Note, this 'parent package' functionality only currently works when the 'child p
 ### `MELOS_PACKAGES`
 
 Define a comma delimited list of package names that Melos should focus on. This bypasses all filtering flags if defined.
+
+### `MELOS_SDK_PATH`
+
+The path to the Dart/Flutter SDK to use. This environment variable has precedence over the `sdkPath`
+option in `melos.yaml`, but is overridden by the command line option `--sdk-path`.
--- a/packages/melos/lib/melos.dart
+++ b/packages/melos/lib/melos.dart
@@ -10,6 +10,7 @@ export 'src/commands/runner.dart'
         ListOutputKind;
 export 'src/common/exception.dart' show CancelledException, MelosException;
 export 'src/common/validation.dart' show MelosConfigException;
+export 'src/global_options.dart' show GlobalOptions;
 export 'src/package.dart' show Package, PackageFilter, PackageMap, PackageType;
 export 'src/workspace.dart' show IdeWorkspace, MelosWorkspace;
 export 'src/workspace_configs.dart'
--- a/packages/melos/lib/src/command_runner.dart
+++ b/packages/melos/lib/src/command_runner.dart
@@ -54,7 +54,8 @@ class MelosCommandRunner extends CommandRunner<void> {
       globalOptionSdkPath,
       help: 'Path to the Dart/Flutter SDK that should be used. This command '
           'line option has precedence over the `sdkPath` option in the '
-          '`melos.yaml` configuration file. If the special value "auto" is '
+          '`melos.yaml` configuration file and the `MELOS_SDK_PATH` '
+          'environment variable. If the special value "auto" is '
           'given, the `sdkPath` setting in `melos.yaml` is ignored and '
           'the SDK is determined from the environment.',
     );
--- a/packages/melos/lib/src/command_runner.dart
+++ b/packages/melos/lib/src/command_runner.dart
@@ -55,9 +55,8 @@ class MelosCommandRunner extends CommandRunner<void> {
       help: 'Path to the Dart/Flutter SDK that should be used. This command '
           'line option has precedence over the `sdkPath` option in the '
           '`melos.yaml` configuration file and the `MELOS_SDK_PATH` '
-          'environment variable. If the special value "auto" is '
-          'given, the `sdkPath` setting in `melos.yaml` is ignored and '
-          'the SDK is determined from the environment.',
+          'environment variable. To use the system-wide SDK, provide '
+          'the special value "auto".',
     );
 
     addCommand(ExecCommand(config));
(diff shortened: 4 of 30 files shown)


blaugold:
Yeah, and some tests are failing too. I'll take look what's going on here.

acoutts:
I have one failed test which seems to be from flutter 2.13 where there are several breaking changes, in this case in HttpClient where they added new overrides.

```
00:00 +10 -1: loading test/package_test.dart [E]                                                                                                                                                                      
  Failed to load "test/package_test.dart":
  ../../../../.pub-cache/hosted/pub.dartlang.org/nock-1.2.0/lib/src/overrides.dart:20:7: Error: The non-abstract class 'MockClient' is missing implementations for these members:
   - HttpClient.connectionFactory=
   - HttpClient.keyLog=
  Try to either
   - provide an implementation,
   - inherit an implementation from a superclass or mixin,
   - mark the class as abstract, or
   - provide a 'noSuchMethod' implementation.
  
  class MockClient implements HttpClient {
        ^^^^^^^^^^
  ../../../flutter/bin/cache/dart-sdk/lib/_http/http.dart:1554:12: Context: 'HttpClient.connectionFactory=' is defined here.
    void set connectionFactory(
             ^^^^^^^^^^^^^^^^^
  ../../../flutter/bin/cache/dart-sdk/lib/_http/http.dart:1704:12: Context: 'HttpClient.keyLog=' is defined here.
    void set keyLog(Function(String line)? callback);
             ^^^^^^
00:01 +107 -2: test/commands/run_test.dart: script supports passing package filter options to "melos exec" scripts [E]                                                                                                
  ScriptException: The script test_script failed to execute
  package:melos/src/commands/run.dart 37:7  _RunMixin.run
  ===== asynchronous gap ===========================
  test/commands/run_test.dart 57:9          main.<fn>.<fn>
  
00:04 +142 ~4 -2: Some tests failed.                        
```

That package needs to release a new major version to support flutter 2.13 and onward.

Otherwise i fixed the other failing test with b006a68ceeb3a934380c0518471996d4fd28867f and that resolved my issue.

acoutts:
This is the assumption breaking the flutter logic:
```
///  c) a lib/main.dart file exists in the package.
```

Not all flutter apps have to use main.dart and in fact it can be whatever the developer wants and can pass to the flutter command via the target parameter.
You could also have a module/package that depends on flutter but has no main.dart because it's intended to be a package and not ran directly.

acoutts:
How strange, even though I set this in my melos.yaml:
```
runPubGetInParallel: false
```

I still had the race condition issues. The only way I could make it work again for me was to override it to return false from `canRunPubGetConcurrently`.

acoutts:
Doh! I did not have my setting yaml structured correctly. Turning off parallelism fixed mine, and ad3e80e fixes the flutter package identification issue.

blaugold:
> Not all flutter apps have to use main.dart and in fact it can be whatever the developer wants and can pass to the flutter command via the target parameter.
You could also have a module/package that depends on flutter but has no main.dart because it's intended to be a package and not ran directly.

The assumption that a Flutter app package always has a `lib/main.dart` is not completely sound, but it's true in most cases. If we don't check for that we get false positives for Flutter packages that are not plugins. You can alwasy just add an empty `lib/main.dart` to your Flutter app. I'd rather not change that in this PR, because it might break something in unexpected ways.

> There seems to be something wrong with how the global option for sdkPath was implemented, digging into it now.

I changed the value that signals that the user wants to use the SDK on the PATH from the empty string to "auto". Thats probably cause the confusion. Using the empty string to convey special meaning is an anti-pattern, IMHO. A meaningful value makes the command line invokation more readable. I don't think 9635af440ccc6651f4854518d85d1592e9f5fd97 and b006a68ceeb3a934380c0518471996d4fd28867f are neccessary if you use `melos <command> --sdk-path auto`.

acoutts:
The issue seems to be in how globalOptions works when we use the melos commands in the melos.yaml.

If I run this command:
```
flutter pub run melos --sdk-path '' exec dart analyze
```

Then `_parseGlobalOptions` runs once and sets sdk path accordingly:
```
[_parseGlobalOptions MelosCommand.sdkPath: ''
```

If I put the same `dart analyze` command above in melos.yaml as a command and run that, now the global options get computed twice - the second time it misses the sdk path setting.

```
flutter pub run melos analyze-ci --sdk-path ''        
[_parseGlobalOptions MelosCommand.sdkPath: ''
melos run analyze-ci
   └> flutter pub run melos exec -- "dart analyze"
       └> RUNNING

[_parseGlobalOptions MelosCommand.sdkPath: 'null'
melos.yaml: SDK path is not valid. Could not find dart tool at /Users/andrew.coutts/Projects/app/.fvm/flutter_sdk2/bin/dart
```

I'm not sure how to fix this now, will wait for your input.

blaugold:
> So i guess the global settings don't cascade down. 

That's actually a good point. There is a good case for adding the `MELOS_SDK_PATH` environment variable. This might be useful in CI, where you can set it globally and then don't have to pass `--sdk-path` everywhere, and we could use it to cascade the SDK path in `melos exec`. 

If an SDK path is specified, we should also add it to the path when running scripts, so that when `dart` or `flutter` is used in a script it uses the SDK.

blaugold:
@acoutts I've added the `MELOS_SDK_PATH` env var and PATH is now updated if an SDK path has been provided. Tests are all passing now, too. Does the PR look good to go for you?

acoutts:
Looks awesome! Thanks a lot for your help.
Only comment I have is i did some testing and noticed the order of precedence is
1. command sdk path (`--sdk-path ..`)
2. `MELOS_SDK_PATH`
3. `melos.yaml` `sdkPath` setting

Do you think it's alright that the command line will override the environment variable, or should it be the other way around? That way if people utilize the command line flag, they can set the flag in CI and they won't have to modify their commands.

blaugold:
My gut feeling is that command line arguments should generally have the highest precedence because they are the most granular way for specifying an option. It's what most tools do, so it's also what most devs expect. It might make sense in this case, but it's not obvious to me in what situation. Do you have a concrete use case?

acoutts:
No other concrete example. You’re right here as well, if someone wants to use the environment variable they could pass the command like argument and check the environment variable in a shell command.

blaugold:
@Salakar This PR is ready to review, when you get a chance.

Salakar:
> @Salakar This PR is ready to review, when you get a chance.

Sorry I'm a bit busy with Google I/O things at the moment 😫  I'm happy for this to go without my review unless you explicitly want me to review?

blaugold:
👍  No problem. I'm fine with merging it without another review.

acoutts:
@blaugold after bumping to melos 2.2.0, my commands that worked on my fork with `--sdk-path auto` do not work now. Is there a change in how the auto api works?

The commit that works for me:
https://github.com/bottlepay/melos.git
`b006a68ceeb3a934380c0518471996d4fd28867f`

To test:
* Set `sdkpath` in `melos.yaml` to some invalid path
* Run a melos command and pass `--sdk-path auto`, it should then use your path to find the flutter sdk and the command will work

If i try the exact same thing on 2.2.0/2.3.0, it instead tries to use the `sdkPath` setting in `melos.yaml` even though i passed "auto" to the command line.

blaugold:
When I set an invalid `sdkPath` in `melos.yaml`, running `melos bootstrap --sdk-path auto`, for example, works as expected for me. What is the full command that is not working?

acoutts:
Aha you're right, so to further explain,
* It works if i run a `melos exec --sdk-path auto ...` from the command line
* It does not work if I define a command in melos.yaml and then try to run it.


melos.yaml command:
```yaml
analyze-ci: flutter pub run melos exec --sdk-path auto -- "dart analyze"
```

Running this will fail:
```
fvm flutter pub run melos analyze-ci
```

But going into the terminal and pasting that command _will_ work.

So the regression between my last commit on my fork and what got merged, is that it's broken for commands put in the melos.yaml file.

blaugold:
You're right, there was a change in `melos run` from the commit you're referencing and the final merged commit. In the final commit, the sdk path is validated when melos starts up before any melos command is executed, which was not the case for `melos run` in the earlier commit. The idea is that Melos always has a valid sdk path and fails fast if that is not the case.

Do you have to work around a limitation where the `MELOS_SDK_PATH` environment variable cannot be used in CI?
Otherwise, you could just have one `analyze` script, which is also simpler:
```
analyze: melos exec -- "dart analyze"
```

acoutts:
I see - that's a good rationale there I agree.
I tried to pass an empty path to the environment variable but it seems to ignore it and still say it doesn't work.
If I try adding `--sdk-path auto` to the melos run command, it also still tries to use `sdkPath` in the melos yaml instead.

🤔

blaugold:
#296 fixes the issue with the `--sdk-path` command line option so that it works for all Melos commands.

acoutts:
works!! thank you so much for your help.

PR #290, opened 2022-05-09
feat: remove dependency overrides `pubspec_overrides.yaml` in `melos clean`

blaugold:
## Type of Change

- [ ] 🗑️ `chore` -- Chore


packages/melos/lib/src/commands/clean.dart
```diff
@@ -40,7 +40,19 @@ mixin _CleanMixin on _Melos {
...
+    // Remove any Melos generated dependency overrides from
+    // `pubspec_overrides.yaml`.
+    final pubspecOverridesFile =
+        File(join(package.path, 'pubspec_overrides.yaml'));
+    if (pubspecOverridesFile.existsSync()) {
+      final contents = await pubspecOverridesFile.readAsString();
+      final updatedContents = mergeMelosPubspecOverrides({}, contents);
+      if (updatedContents != null) {
+        await pubspecOverridesFile.writeAsString(updatedContents);
       }
```
Salakar, line 56: I'm not sure but do we want to delete the file here in an `else` since it's empty and since it was most likely created by Melos?

blaugold: `mergeMelosPubspecOverrides` allows overrides that don't come from melos, so passing in `{}` does not guarantee that `pubspec_overrides.yaml` is empty. But we can check for that and if there are no other overrides, then I think it's a good idea to delete the file.

blaugold: Done

Salakar (approved):
Amazing, thanks for this :)

PR #300, opened 2022-05-19
fix: Run Process.runSync in another shell to get Dart version

lulupointu:
## Description

When trying to get the dart version, `dart --version` should be run inside a separate shell to avoid failing when running in a sub process.

The failing of this command is actually used here:
https://github.com/invertase/melos/blob/431ecb275bf7169d71f9c19e4593abd3274a46ca/packages/melos/lib/src/common/utils.dart#L404-L411

This bug was introduced in https://github.com/invertase/melos/commit/740050c4dd67938d0674ddd37f0291d52f331bd4 since it changed how to dart version is retrieved. 

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/common/utils.dart
+++ b/packages/melos/lib/src/common/utils.dart
@@ -120,6 +120,7 @@ Version currentDartVersion(String dartTool) {
     ['--version'],
     stdoutEncoding: utf8,
     stderrEncoding: utf8,
+    runInShell: true,
   );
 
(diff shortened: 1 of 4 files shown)


blaugold:
Thanks for the PR! I don't quite understand yet what the behavior is that this fixes. Do you get an error and if so, what is it?

Salakar (approved):

PR #320, opened 2022-06-09
ci: Use GitHub actions

spydon:
## Description

This makes use of GitHub actions for the pipeline instead of separate scripts for each platform.

## Type of Change

- [x] ✅ `ci` -- Build configuration change


blaugold:
In case you're wondering, I think the test failures have to do with changing how melos is installed.
The tests expect that the local version of melos is also on the path, which can be accomplished with running this from the repo root:
```
dart pub global activate --source="path" . --executable="melos"
```

spydon:
> In case you're wondering, I think the test failures have to do with changing how melos is installed. The tests expect that the local version of melos is also on the path, which can be accomplished with running this from the repo root:
> 
> ```
> dart pub global activate --source="path" . --executable="melos"
> ```

Ah that makes sense, didn't think about that this can't test itself with the melos action.
I'll add that to the pipeline instead.

Salakar (approved):
Amazing, thanks for another CI PR :)

PR #323, opened 2022-06-15
feat: Add --diff filter

olexale:
## Description

We found it extremely helpful to have the possibility to run melos on packages that are different between branches. As the `since` option does not really work with the difference I propose to introduce a `diff` option that will call `git diff` instead of `git log`.

The usage might be:
```sh
melos list --diff=some_branch
```
or
```sh
melos list --diff=origin/main...HEAD
```

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,11 +40,11 @@ To solve these (and many other) problems, some projects will organize their code
 To setup and use this melos mono repo locally for the purposes of contributing, clone it and run the following commands from the root of the repository:
 
 ```bash
-# Remove previous instances of melos:
-dart pub global deactivate melos
+# Install melos if it's not already installed:
+dart pub global activate melos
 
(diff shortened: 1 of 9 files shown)


blaugold:
Thank you for your contribution! 🎉 

Could you add documentation for the new option in the [docs](https://github.com/invertase/melos/blob/main/docs/filters.mdx), with examples for a simple ref and ranges? 

@Salakar
`--diff=<ref>...HEAD` is equivalent to `--since=<ref>`, so we could deprecate `--since` in favor of `--diff`.

olexale:
@blaugold, done, please review.

blaugold:
I've made `diff` and `since` mutually exclusive. Otherwise, LGTM.

PR #326, opened 2022-06-18
feat: add runPubGetOffline option

mspb1g12:
## Description

Add an option to run all pub gets offline, relying on pre-populated pub caches.
This is useful in our CI environment which has a locked down network. Ideally the `--offline` flag (and other pub get flags) should be passed from `bootstrap` through to `pub get` but this seems like the simpler change in the short run, though if that seems preferable I can look into how it might work.
Another alternative could be `runPubGetOfflineInCI` using https://pub.dev/packages/ci to detect the CI environment which would be more useful but may be too specific a use-case.

## Type of Change

- [ ] 🗑️ `chore` -- Chore


blaugold (approved):
Thanks for contributing! LGTM

PR #349, opened 2022-07-10
feat: Add support for specifying an IntelliJ module name prefix

ryanhanks:
## Description

This change came about as a result of [discussion around the 'melos_' string](https://github.com/invertase/melos/discussions/336) that is prepended to IntelliJ project module names, and the desire for that prefix to be blank.

The IntelliJ configuration parser currently assumes a single boolean value, so the parser has been updated to now support a map with which to specify intelliJ config.

The default value of enabled can now be overrode via an `enabled` key in the intellij map, and the parser continues support the current configuration approach, whereby the `intellij` key is set to a boolean value:
```yaml
ide:
  intellij:
    enabled: false  # this is now the documented approach
```
```yaml
ide:
  intellij: false        # the config parser will support this schema also
```

This map now allows for the specification of an alternative project module name prefix:
```
ide:
  intellij:
    moduleNamePrefix: ''   # set to empty string to override default value of 'melos_'
```

## Type of Change

## Notes
* I added docs around the current integration with IntelliJ to the best of my knowledge. Let me know what I missed.


ryanhanks:
Also, I didn't update the current `melos.yaml` config for this project, and I thought it might break CI. Lmk if we should include a change for that in this PR or if it can / should be addressed at another time.

ryanhanks:
Looks like I missed a couple things here. Updates to follow in the next day or few.

blaugold:
Thanks for this contribution! Looks great, aside from the analyzer errors (just some style issue) and formatting.

> Also, I didn't update the current melos.yaml config for this project, a I thought it might break CI. Lmk if we should include a change for that in this PR or if it can / should be addressed at another time.

I think it's fine to leave it as is for now.

ryanhanks:
Thanks blaugold 👍 

For transparency, in addition to the style cleanup, I missed tracing the "melos_" prefix all the way down the stack, and it looks like it exists in hard-coded form beyond the areas I've addressed thus far, e.g., https://github.com/invertase/melos/blob/70448bec7d8cf5f8d0a8dc8c2660e70033936329/packages/melos/lib/src/common/intellij_project.dart#L82-L84 (I also noticed this value is hard-coded in the clean-up script)

blaugold:
Hey @ryanhanks, I've rebased the PR and added the use of `moduleNamePrefix` in all places where module names are built.
For the run configurations, we should keep using the hard-coded prefix.
I think it's ready to be merged. WDYT?

ryanhanks:
Hey @blaugold 

I pulled it down and gave it a run this morning, looks good to me!

I must say, on my repo of 26 projects, it was v nice to open IntelliJ and see the results of this work :)

Thanks for cleaning this up! Feel free to merge whenever you're ready.

Salakar (approved):
Nice!

PR #359, opened 2022-08-08
fix: `scripts.*.exec.failFast` option in `melos.yaml`

alex-sandri:
## Description

`--fail-fast` is a flag and not an option and as such does not accept any value.

### Example

This config is currently broken as in the generated command `--fail-fast true` is not valid syntax.

```yaml
scripts:
  lint:
    run: flutter analyze
    exec:
      failFast: true
      concurrency: 1
```

## Type of Change

- [ ] 🗑️ `chore` -- Chore


blaugold:
Good catch! Thank you.

PR #405, opened 2022-10-13
feat: allow help to be shown from anywhere

### LINKED ISSUE #376: Allow viewing usage description with `-h` when outside a Melos workspace
Currently, Melos always reports an error when invoked outside a Melos workspace. To explore Melos commands, it's useful to be able to view the usage description outside a Melos workspace.

SscSPs:
Closes #376

Signed-off-by: Sahil Soni <SscSPs@gmail.com>

## Description
Allow help to be run from anywhere and not just from inside a melos workspace

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/bin/melos.dart
+++ b/packages/melos/bin/melos.dart
@@ -40,7 +40,13 @@ Future<void> main(List<String> arguments) async {
     return;
   }
   try {
-    final config = await MelosWorkspaceConfig.fromDirectory(Directory.current);
+    var allowOutside = false;
+    if (arguments.contains('--help') || arguments.contains('-h')) {
+      allowOutside = true;
+    }
+    final config = allowOutside
+        ? MelosWorkspaceConfig.empty()
+        : await MelosWorkspaceConfig.fromDirectory(Directory.current);
     await MelosCommandRunner(config).run(arguments);
   } on MelosException catch (err) {
     stderr.writeln(err.toString());
--- a/packages/melos/bin/melos.dart
+++ b/packages/melos/bin/melos.dart
@@ -40,11 +40,7 @@ Future<void> main(List<String> arguments) async {
     return;
   }
   try {
-    var allowOutside = false;
-    if (arguments.contains('--help') || arguments.contains('-h')) {
-      allowOutside = true;
-    }
-    final config = allowOutside
+    final config = shouldUseEmptyConfig(arguments)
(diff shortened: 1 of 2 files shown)


SscSPs:
Can you please review this PR
And, does this count under hacktoberFest? just making sure, I think it should, but not sure, added the closes tag with a hacktoberfest issue, do i need to add something else as well ?

packages/melos/bin/melos.dart
```diff
@@ -40,7 +40,13 @@ Future<void> main(List<String> arguments) async {
     return;
   }
   try {
-    final config = await MelosWorkspaceConfig.fromDirectory(Directory.current);
+    var allowOutside = false;
+    if (arguments.contains('--help') || arguments.contains('-h')) {
+      allowOutside = true;
+    }
```
blaugold, line 46: This can just be:

```dart
final hasHelpOption = arguments.contains('--help') || arguments.contains('-h');
```

blaugold (changes requested):
Thanks for this contribution!

LGTM. Just one nit.

packages/melos/bin/melos.dart
SscSPs: The reason i put it this way was so if in future we want more flags to be available outside the melos workspaces, we can update this condition here.

blaugold: How about this, then:
```dart
final hasHelpOption = arguments.contains('--help') || arguments.contains('-h');
final useEmptyConfig = hasHelpOption;
```
We avoid a mutable variable, it's shorter. If we later want to run with an empty config in other situations, we can add `|| <condition>` behind `hasHelpOption` and have an easily readable list of conditions when we need to use an empty config.

SscSPs: Makes sense, may be as well just move it out to a function, 
```
bool shouldUseEmptyConfig(List<String> arguments)
```
And move all this there and use that

blaugold: That sounds good.

SscSPs: Cool, will make the changes tomorrow, busy today

SscSPs: made the changes, let me know if there's something else needed as well :)

blaugold (approved):

Issue #258, opened 2022-03-07
request: Configurable prefix before conventional commit title


spydon:
### Is there an existing feature request for this?

- [X] I have searched the existing issues.

### Description

In some collaborative version control services, like Azure DevOps, they add an annoying non-configurable prefix to each merged pull requests commit message. 

For example: `Merged PR 404: fix: Use correct path for signing`

This feature would implement so that it is possible to ignore that prefix, maybe with a regex, to be able to version melos projects that are hosted by services with these prefixes properly.

### Reasoning

It would be useful since it would be possible to use Melos in services that are using these kind of prefixes.

### Additional context and comments

[Feature request for Azure Devops](https://developercommunity.visualstudio.com/t/change-default-title-for-pull-request-commits-to-n-1/365716)

Salakar:
> @Salakar could I work on a PR for this? If that is the case, could you give some pointers to how you'd want it implemented? 🙂

Of course! Best place to start would be the conventional commits package: https://github.com/invertase/melos/tree/main/packages/conventional_commit - if the goal is to just ignore the prefix then it should be pretty straightforward modifying the regex to handle it. If it needs to be configureable then could probably add it as a optional named arg to `tryParse`: https://github.com/invertase/melos/blob/main/packages/conventional_commit/lib/conventional_commit.dart#L75 . Package has pretty good test coverage so should be hard to break things without it being detected.

Salakar:
I mean technically removing the `^` from the start of the regex also solves this problem :D https://github.com/invertase/melos/blob/main/packages/conventional_commit/lib/conventional_commit.dart#L19

spydon:
> I mean technically removing the `^` from the start of the regex also solves this problem :D https://github.com/invertase/melos/blob/main/packages/conventional_commit/lib/conventional_commit.dart#L19

Hahah here I was imagining a few hours of work at least. 😅 
That sounds like a good solution to me at least!

as19ish:
Can we add support for bitbucket ? In bitbucket also when we try to merge PR it add this message as default.
Merged in feature/DM-175-tag-management (pull request #156)

Issue #24, opened 2020-11-05
melos version (feature request)


lesnitsky:
As a maintainer of a monorepo I'd like to have a possibility to bump package versions with help of melos tool. Please let me know if that's in scope of the responsibilities of melos

lesnitsky:
oops, that's already exist :D

Salakar:
@lesnitsky forgot to reply to this but this has been now implemented :)

lesnitsky:
awesome, thanks!

PR #407, opened 2022-10-14
fix: Merge commits should be versioned

lohnn:
## Description

Merge commits were not versioned. Example:

```Merged PR 123: fix(3305): change url launcher to open externally```

Fixes a regression of #258.

Me and @spydon fixed this together :)

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/conventional_commit/test/conventional_commit_test.dart
+++ b/packages/conventional_commit/test/conventional_commit_test.dart
@@ -93,13 +93,24 @@ void main() {
       'correctly parses commit with prefix before conventional commit type',
       () {
         const commitMessage = 'Merged PR 404: feat(scope): new feature';
-        final conventionalCommit = ConventionalCommit.tryParse(commitMessage);
-        expect(conventionalCommit?.type, 'feat');
-        expect(conventionalCommit?.scopes, ['scope']);
-        expect(conventionalCommit?.description, 'new feature');
-        expect(conventionalCommit?.isMergeCommit, true);
+        final conventionalCommit = ConventionalCommit.tryParse(commitMessage)!;
+        expect(conventionalCommit.type, 'feat');
+        expect(conventionalCommit.scopes, ['scope']);
+        expect(conventionalCommit.description, 'new feature');
+        expect(conventionalCommit.isMergeCommit, true);
       },
     );
+
+    test('parses merge commits which are not conventional commits', () {
+      const commitMessage = 'Merged foo into bar';
+      final conventionalCommit = ConventionalCommit.tryParse(commitMessage)!;
+      expect(conventionalCommit.type, isNull);
+      expect(conventionalCommit.scopes, isEmpty);
+      expect(conventionalCommit.description, isNull);
+      expect(conventionalCommit.body, isNull);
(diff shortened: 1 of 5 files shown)


blaugold:
I think there is a bit of inconsistency in the current code which we need to resolve to make this fully work.

In `conventional_commit` we are currently parsing merge commits as a `ConventionalCommit` even if it does not have the correct header (`fix: ...`). Any commit that starts with `Merge ` is allowed and gets some defaults.

https://github.com/invertase/melos/blob/9ea87a3252ee07592f145bd2212b1015fd714168/packages/conventional_commit/lib/conventional_commit.dart#L74-L84

This is not part of the Conventional Commit spec. I don't think we should do that and return `null` in that case, since it's not an actual Conventional Commit. We will need to know that every `ConventionalCommit` is an actual Conventional Commit for this PR.

There are also two places in the changelog generation that need to be updated. `_filteredAndSortedCommits` currently filters out all merge commits:

https://github.com/invertase/melos/blob/9ea87a3252ee07592f145bd2212b1015fd714168/packages/melos/lib/src/common/changelog.dart#L213-L214

And merge commits get special cased (even though they will always be filtered out before 🤷), which they shouldn't be with the changes in this PR:

https://github.com/invertase/melos/blob/9ea87a3252ee07592f145bd2212b1015fd714168/packages/melos/lib/src/common/changelog.dart#L173-L176

lohnn:
Looking around, it seems a the way people found around this problem was to add a non-capturing group to the input to the header regex at runtime. Such as [this](https://github.com/conventional-changelog/standard-version/issues/690#issuecomment-1036672116).
Would the proper way of solving this issue here be to add support for header regex override?

blaugold:
I think directly supporting prefixes like `Merged PR 123:` is good, so users don't have to do extra set up.

My concern is that we distinguish `Merged PR #24: fix: ...` (which is a valid, versionable conventional commit), from `Merged foo into main`. Both parse as `ConventionalCommit`s, but only the former commit has a `type` and that is what we test in `isVersionableCommit`, so that's fine.

I've added some tests and added the changes to fix changelog generation.

lohnn:
Thanks a lot 🙏 Happy to see even more tests are added due to this PR as well.

Anything else you want me to do? Happy to help if there is.

lohnn:
That commit message does not follow conventional commits. It has to include at least a `<type>:`. in the commit message, else it won't be able to parse as a conventional commit.

as19ish:
@lohnn yes we follow conventional commit but when we try to merge bitbucket add some text from there end.

Screenshot 2022-10-19 at 7 50 43 PM

blaugold:
@as19ish Could you open a new issue with details about how you merge commits (rebase, squash, merge), what bitbucket adds and how this breaks versioning?

Let's handle that separately, since this PR is about fixing a regression.

blaugold (approved):

spydon:
Thanks for merging @blaugold!
Any chance we could have a release including this? 😇

blaugold:
I think we have a bunch of changes people are looking forward to. @Salakar Could you publish a new release?

PR #421, opened 2022-11-17
feat: implement scoped changelogs

lesnitsky:


lesnitsky:
for the context:

in [flutterfire](https://github.com/firebase/flutterfire) we have a bunch of `firebase_ui_*` packages and we want to have a separate changelog for all packages that start with this prefix.

blaugold (approved):
Looks basically good. I've cleaned up a few things, made CI pass and added docs.

Salakar (approved):

docs/configuration/overview.mdx
```diff
@@ -259,6 +261,38 @@ command:
...
+        scope: foo_*
+        description: |
+          All notable changes to foo packages will be documented in this file.
+```
+
+### `command/version/changelogs/path`
+
+The path to the changelog file relative to the workspace root.
+
+### `command/version/changelogs/scope`
```
Salakar, line 283: Do you think maybe we should have the same syntax as `select-package` in melos run scripts

Will allow all filter types then rather than just scope,

Should just be able to reuse most of the logic to do so?

Thoughts @blaugold?

blaugold: Good point! I think we should do that.

PR #431, opened 2022-12-08
feat!: local installation of melos in workspace

blaugold:
Before this change, versioning of Melos by users was not enforced. Most users installed Melos globally, and without specify a version, on development as well as CI machines. That means potentially breaking versions of Melos were adopted without the users' knowledge.

With this change, users are required to add a dependency on `melos` to a `pubspec.yaml` file, next to the `melos.yaml` file. The globally installed `melos` executable will hand over to the locally installed version if their versions are different. This ensures that the version of Melos used is always the correct one for the workspace the user is currently working in.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/.github/workflows/scripts/install-tools.bat
+++ b/.github/workflows/scripts/install-tools.bat
@@ -1,2 +1,4 @@
 CMD /C dart pub global activate --source=path . --executable=melos --overwrite
+REM Workaround an issue when running global executables on Windows for the first time.
+CMD /C melos > NUL
 melos bootstrap
--- a/.github/workflows/scripts/install-tools.bat
+++ b/.github/workflows/scripts/install-tools.bat
@@ -1,4 +1,2 @@
-CMD /C dart pub global activate --source=path . --executable=melos
-REM Workaround an issue when running global executables on Windows for the first time.
-CMD /C melos > NUL
+CMD /C dart pub global activate --source=path . --executable=melos --overwrite
 melos bootstrap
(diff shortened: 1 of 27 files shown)


Ehesp:
Just to be clear here - it'll use local melos if installed, if not fallback to global melos?

oravecz:
I 'm surprised that the tool would put restrictions on how I structure my project.
I think my use case is far more common because developers generally start with the `App` in `/lib` and branch out into `/packages/*` as the project grows.

> But we should generally use the lowest possible version bound and fix it if that is not the case.

Sorry, not sure what this means. Will you be relying on path@^1.8.2 instead of path@^1.8.3 in the next pre-release?

blaugold:
I've added a test to our CI to ensure `melos` can be installed alongside Flutter dependencies on the stable channel.

The dependency on `path@1.8.3` came from a dependency of `melos`. This package has lowered its version constraint on `path`, so `melos` can now be installed alongside Flutter without requiring a new release of `melos`.

> I think my use case is far more common because developers generally start with the App in /lib and branch out into /packages/* as the project grows.

Thanks for the feedback!  The focus of Melos was mostly development of packages instead of apps, but that seems to shift, and we certainly want to support both use cases.

RemyTabardel:
@blaugold How to manage multi-clients application ?

I had that, and I was running melos on all packages no matter the client (flutter get, clean, build runner ...) :

- root
- root/melos.yaml
- root/business/common-business-1/pubspec.yaml
- root/business/common-business-2/pubspec.yaml
- root/apps/client1/pubspec.yaml (entry point, application)
- root/apps/client2/pubspec.yaml (entry point, application)

I need to add a fake pubspec.yaml under root ?

PR #440, opened 2022-12-27
feat: support running scripts in topological order with `melos exec`

### LINKED ISSUE #433: request: run `melos exec` in different packages in topologically order

### Version

2.7.0

### Description

Melos creates the proper graph, but when I execute my "generate" command, it doesn't begin generation with the graph's leaf nodes before working its way up the tree. In my case, I am building `json_serializable` and `freezed` classes and dependencies between projects are messed up because the parent is built before the child.

I've removed concurrency from my `generate` command to ensure that wasn't the issue.

Example of generate command.

```yaml
  generate:
    run: melos exec -c 1 --fail-fast -- flutter pub run build_runner build --delete-conflicting-outputs
    description: Runs any code generation controlled by the build_runner script
    select-package:
      depends-on: build_runner
```

### Steps to reproduce

I don't have a reproducible step. I'm not even sure that Melos is smart enough to call packages in the correct order, so this may be a feature request and not a bug. Seems pretty fundamental for a tool like this, though.

### Expected behavior

I expect Melos to generate a tree structure of all package dependencies, and run all leaf nodes in parallel while traversing up the tree. Melos has the proper graph data, but without this kind of logic, an option to declare build order would be acceptable.

### Screenshots

_No response_

Currently, I workaround the problem by moving my `generate` script to `postgenerate` and manually running `generate` on the package that seems to be out of order in the `generate` script.

mspb1g12:
## Description

Add `--require-dependencies` flag to exec to require the dependencies of a package be exec'd before the package itself.
As part of this, reorder the exec's to be topological like publish.

I believe this should close #433 

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/docs/configuration/scripts.mdx
+++ b/docs/configuration/scripts.mdx
@@ -91,10 +91,12 @@ Whether `exec` should fail fast and not execute the script in further packages i
 in an individual package.
 Defaults to `false`.
 
-## `scripts/*/exec/requireDependencies`
+## `scripts/*/exec/orderDependents`
 
-Whether `exec` should require the dependencies of this package to be completed first. Only applies
-to dependencies that would normally be part of this command.
+Whether `exec` should order the execution of the script in multiple packages based on the
+dependency graph of the packages. The script will be executed in leaf packages first and then
+in packages that depend on them and so on. This is useful for example, for a script that generates
+code in multiple packages, which depend on each other.
 Defaults to `false`.
 
 ## `scripts/*/env`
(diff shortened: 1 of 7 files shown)


OPVL (approved):

jrcryer (approved):

blaugold (approved):
Thanks for this awesome contribution! I've cleanup a few small issues and renamed the option to make it a bit more self-explanatory and use the same terminology that other options use.

PR #468, opened 2023-02-07
fix: use `git pull --tags` instead of `git fetch --tags`

russellwheatley:
## Description

Use `git pull --tags` instead of `git fetch --tags`. The problem with using `fetch` - the tag needs to be associated with a branch that is on remote. If it isn't (i.e. usually squashed and merge) it will not be retrieved and will remain orphaned on remote.

Using `git pull --tags` will pull all tags no matter the status of the branch.

## Type of Change

- [X] 🛠️ `fix` -- Bug fix (non-breaking change which fixes an issue)
- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/common/git.dart
+++ b/packages/melos/lib/src/common/git.dart
@@ -229,7 +229,7 @@ Future<void> gitFetchTags({
   required MelosLogger logger,
 }) async {
   await gitExecuteCommand(
-    arguments: ['fetch', '--tags'],
+    arguments: ['pull', '--tags'],
     workingDirectory: workingDirectory,
     logger: logger,
   );


Salakar (approved):

blaugold:
I couldn't find anything on this difference between `fetch` and `pull` in the git docs. The docs for `pull` even say the following:

> More precisely, git pull runs git fetch with the given parameters and then depending on configuration options or command line flags, will call either git rebase or git merge to reconcile diverging branches.

Is this difference documented anywhere? I'd love to understand it better.

If we need to use `git pull` we probably need to be a bit more careful. Usually the current branch should be up to date with the remote before running `melos version` but users can make mistakes. I think we should require that if the current branch has a remote tracking branch it is up to date at the beginning of `melos version` and abort if it is not. Otherwise the behaviour is a bit unpredictable for users.

Salakar:
We've had cases on FlutterFire where `git fetch --tags` was run but tags that are not attached to a ref weren't being pulled and so older tags were being used to generate the changes, I'm not exactly sure why though, using pull seemed to do the trick at the time. 

For context, I asked Russell to post this PR based on the FF release process; https://github.com/firebase/flutterfire/blob/master/CONTRIBUTING.md#run-a-release

spydon (approved):
Lgtm! I think this change seems sensible after reading up on `git pull --tags` vs `git fetch --tags`

PR #525, opened 2023-05-16
fix: allow passing additional arguments to multiline run commands

### LINKED ISSUE #232: [bug] `melos run` args only work in single line scripts
Hi there! 👋🏻 

@Salakar graciously added the functionality to run `melos run` scripts with additional arguments in #231. 😄

### Description

Looking at the tests and trying it out myself, it works fine in single line scripts.

```yaml
// melos.yaml

name: test_args

packages:
  - packages/**

scripts:
  testargs:
    run: echo "$0" "$1" "$2" "$3"
```

Output:

```shell
$ melos run testargs first second third
melos run testargs
   └> echo "$0" "$1" "$2" "$3"
       └> RUNNING

/bin/sh    first second third

melos run testargs
   └> echo "$0" "$1" "$2" "$3"
       └> SUCCESS
```

### Actual behavior

However, using a multi-line string in YAML to define a script breaks this behavior.

#### Example 1

```yaml
// melos.yaml

name: test_args

packages:
  - packages/**

scripts:
  testargs:
    run: |
      echo "$0" "$1" "$2" "$3"
```

Output:

```shell
$ melos run testargs first second third
melos run testargs
   └> echo "$0" "$1" "$2" "$3"
       └> RUNNING

/bin/sh   
/bin/sh: line 1: first: command not found


melos run testargs
   └> echo "$0" "$1" "$2" "$3"
       └> FAILED
ScriptException: The script testargs failed to execute
```

#### Example 2

```yaml
// melos.yaml

name: test_args

packages:
  - packages/**

scripts:
  testargs:
    run: |
      echo "Shell: $0"
      echo "First: $1"
      echo "Second: $2"
      echo "Third: $3"
```

Output:

```shell
$ melos run testargs first second third
melos run testargs
   └> echo "Shell: $0"echo "First: $1"echo "Second: $2"echo "Third: $3"
       └> RUNNING

Shell: /bin/sh
First: 
Second: 
Third: 
/bin/sh: line 4: first: command not found


melos run testargs
   └> echo "Shell: $0"echo "First: $1"echo "Second: $2"echo "Third: $3"
       └> FAILED
ScriptException: The script testargs failed to execute
```

### Expected behavior

I believe arguments should be handled independently of the format of the script payload provided to Melos, just like when executing a `some_script.sh first second third` would parse `$1`, `$2`, etc. correctly.

---

If you need any additional info or help, feel free to reach out! 😄 
Great job y'all 👍🏻

KoheiKanagu:
## Description

Fix #232

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -156,10 +156,11 @@ mixin _RunMixin on _Melos {
     final scriptParts = scriptSource.split(' ');
 
     return startCommand(
-      scriptParts..addAll(extraArgs),
+      scriptParts,
       logger: logger,
       environment: environment,
       workingDirectory: config.path,
+      extraArgs: extraArgs,
     );
   }
 }
--- a/packages/melos/lib/src/common/utils.dart
+++ b/packages/melos/lib/src/common/utils.dart
@@ -437,6 +437,7 @@ Future<Process> startCommandRaw(
   String? workingDirectory,
   Map<String, String> environment = const {},
   bool includeParentEnvironment = true,
+  List<String> extraArgs = const [],
 }) {
   final executable = currentPlatform.isWindows ? 'cmd.exe' : '/bin/sh';
   workingDirectory ??= Directory.current.path;
@@ -444,8 +445,8 @@ Future<Process> startCommandRaw(
   return Process.start(
     executable,
     currentPlatform.isWindows
-        ? ['/C', '%MELOS_SCRIPT%']
-        : ['-c', r'eval "$MELOS_SCRIPT"'],
+        ? ['/C', '%MELOS_SCRIPT%', ...extraArgs]
+        : ['-c', r'eval "$MELOS_SCRIPT"', ...extraArgs],
     workingDirectory: workingDirectory,
     environment: {
       ...environment,
@@ -464,6 +465,7 @@ Future<int> startCommand(
   bool onlyOutputOnError = false,
   bool includeParentEnvironment = true,
   required MelosLogger logger,
+  List<String> extraArgs = const [],
 }) async {
   final processedCommand = command
       .map((arg) {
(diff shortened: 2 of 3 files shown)


spydon:
@KoheiKanagu sorry for the long delay, is this still needed after #540?

KoheiKanagu:
@spydon Yes, it is needed.
#232 has not been resolved.

spydon:
@blaugold you had some comments about this functionality in the issue, what do you think about this PR?

blaugold:
In its current state, this PR would be a breaking change, because it changes from appending extra args to the script to passing them as extra args to the shell, which means they are only accessible through `$1`, `$2`, ... 

In general, multiline scripts are a bit of a footgun and I don't think we should expand support for them. Melos just passes the script to the system shell. To ensure that multiple commands within a script fail fast, they all have to be concatenated through `&&`. This is the only approach that works across Windows and UNIX system shells.

I think we should keep Melos simple and keep the current behaviour of forwarding scripts with minimal processing to the system shell. This is how NPM scripts work, too. I would recommend maintaining scripts that are more complex in a separate file and just invoke the file through Melos. If you do that, extra parameters can be accessed through `$1`, `$2` (in the case of `sh` type shells) in the script.

If we really want to support multiline scripts, we should look at CI/CD platforms like GitHub Actions for how to deal with cross-platform concerns and how to avoid unexpected behaviour.

KoheiKanagu:
I agree that Melos should be kept simple.

In the current Melos, multiline scripts can be executed, but they may not behave as expected. For example, #232 shows an error that occurs when a multiline script is executed.

I think Melos should return an error when trying to execute a multiline script. Or, it would be better to state in the documentation that multiline scripts are not supported.
This would prevent users from encountering unexpected errors.

spydon:
I'll close this and then we can continue to discuss in the issue, thanks for the effort though!

Issue #94, opened 2021-04-08
Request: common dependencies


isacjunior:
In some cases, we have common dependencies around some packages. For now, we have two workarounds:
- Create a package `common_dependencies` and centralized the common dependencies here.
[✗] The dependency is installed as a `transitive` dependency and we don't have `IntelliSense`.
[✓] All common dependencies in a unique locale. This is easier to manipulate versions.
- Install repeatedly dependencies in each package.
[✗] Is difficult to manipulate the version, needs to change in many locales.
[✓] We have the dependency as `direct` and so we have `IntelliSense`

Any know the workaround to solve this?

## Attempt 
I tried to use the following below but it doesn't work.
> **melos version:** ` 0.4.10+1`

```yaml
# melos.yaml
dependencies:
  http: ^0.13.1
```

## Suggestion:

When the melos copy `pubspec.yaml` and `pubspec.lock`, we could insert these dependencies on the specific packages.

```yaml
# melos.yaml
dependencies:
  http: ^0.13.1
    scope: "package_one,package_two"

dev_dependencies:
  build_runner: ^1.12.2
    ignore: "package_one"
```

Ehesp:
Have you tried to add the dependencies to the root `pubspec.yaml` (not `melos.yaml`)?

isacjunior:
> Have you tried to add the dependencies to the root `pubspec.yaml` (not `melos.yaml`)?
> 
> Then add the root package, like so: https://github.com/invertase/melos/blob/master/melos.yaml#L4

This is the same approach that I mentioned above. The root project would a `common_dependencie` package.

The solution for now that I find is to use [getx_cli](https://github.com/jonataslaw/get_cli)  to install the common packages. But would amazing if `melos` could have this feature.

```yaml
common_dependencies:
  run: melos exec -c 1 --scope="package_one,package_two" -- exec "get install provider:5.0.0 http:0.13.1"
```

SAGARSURI:
This would be great if we can define common dependencies in `melos.yaml`.

SAGARSURI:
Hey @Salakar would it possible to add this feature in Melos?

blaugold:
This is not currently supported by Melos. But if you just want to add `dart_code_metrics` to all packages, you can use `melos exec`:
```shell
melos exec -- dart pub add --dev dart_code_metrics:^4.15.0
```

BenVercammen:
[...]

Issue #472, opened 2023-02-09
request:  add `melos:` section to root pubspec.yaml instead of requiring `melos.yaml` in Melos v3


GregoryConrad:
### Description

Just as the title says, adding a simple `melos` section to the root pubspec.yaml instead of needing a different `melos.yaml` file might be a nice feature, since Melos v3 requires a root pubspec.yaml already.

### Reasoning

It would be nice to keep all monorepo config in the same file, considering the Melos version is going to be set in the pubpsec as well. Other packages (i.e., `flutter_launcher_icons` and many in the NPM world) do this.

PR #526, opened 2023-05-17
feat: Sync common dependency versions between packages

lohnn:
**NOTE:** I'm adding this as a draft PR as docs and tests are not yet added and I haven't yet added feedback if the user turned on the flag but didn't provide the `common_packages.yaml`-file. But I wish for feedback on my solution anyways.

## Description

Added support for setting versions for common/shared dependencies that will be applied to all packages in a Melos monorepo project. Implements #94.

A new bootstrap config flag (`shareDependencies`) is added to start using this new feature, and then the dependency versions for `dependencies` and `dev_dependencies` in `common_packages.yaml` will be applied to all packages that contains said dependency.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/docs/commands/bootstrap.mdx
+++ b/docs/commands/bootstrap.mdx
@@ -89,37 +89,35 @@ melos bootstrap --diff="main"
 Melos bootstrap command supports a few different flags that can be defined in
 your `melos.yaml`.
 
+
 ### Shared dependencies
 
 If you want to share dependency versions between your packages in your Melos
-project, add the flag `runPubGetInParallel` to your bootstrap config and add a
-file `common_packages.yaml` to your project.
+project, just add the dependencies you wish to share between the packages to
+your bootstrap config in your `melos.yaml` file.
+
+If a dependency from `environment`, `dependencies` or `dev_dependencies` in
+your `common_packages.yaml` exists in a package, the dependency version in this
+package will be updated to match the version defined in your
+bootstrap config every time `melos bootstrap` is run.
 
 ```yaml
 # melos.yaml
 # ...
 command:
+  environment:
+    sdk: ">=3.0.0 <4.0.0"
+    flutter: ">=3.0.0 <4.0.0"  
   bootstrap:
-    shareDependencies: true
-# ...
-```
-
-If a dependency from `dependencies` or `dev_dependencies` in your
-`common_packages.yaml` exists in a package, the dependency version in this
-package will be updated to match the version defined in your
-`common_packages.yaml` file every time `melos bootstrap` is run.
-
-```yaml
-# common_packages.yaml
-# ...
-dependencies:
-  collection: ^1.17.0
-  uni_links2:
-  uni_links_macos:
-    git: https://github.com/SamJakob/uni_links_macos.git
-
-dev_dependencies:
-  build_runner: ^2.3.3
+    dependencies:
+      collection: ^1.18.0
+      integral_isolates: any
+      uni_links2:
+      uni_links_macos:
+        git: https://github.com/SamJakob/uni_links_macos.git
+    
+    dev_dependencies:
+      build_runner: ^2.3.3
 # ...
 ```
 
--- a/docs/commands/bootstrap.mdx
+++ b/docs/commands/bootstrap.mdx
@@ -84,6 +84,45 @@ example:
 melos bootstrap --diff="main"
 ```
 
+## Bootstrap flags
+
+Melos bootstrap command supports a few different flags that can be defined in
+your `melos.yaml`.
+
+### Shared dependencies
+
+If you want to share dependency versions between your packages in your Melos
+project, add the flag `runPubGetInParallel` to your bootstrap config and add a
+file `common_packages.yaml` to your project.
+
+```yaml
+# melos.yaml
+# ...
+command:
+  bootstrap:
+    shareDependencies: true
+# ...
+```
+
+If a dependency from `dependencies` or `dev_dependencies` in your
+`common_packages.yaml` exists in a package, the dependency version in this
+package will be updated to match the version defined in your
+`common_packages.yaml` file every time `melos bootstrap` is run.
+
+```yaml
+# common_packages.yaml
+# ...
+dependencies:
+  collection: ^1.17.0
+  uni_links2:
+  uni_links_macos:
+    git: https://github.com/SamJakob/uni_links_macos.git
+
+dev_dependencies:
+  build_runner: ^2.3.3
+# ...
+```
+
 ## Adding a post bootstrap lifecycle script
 
 Melos supports various command [lifecycle hooks](/configuration/scripts#hooks)
--- a/packages/melos/lib/src/commands/bootstrap.dart
+++ b/packages/melos/lib/src/commands/bootstrap.dart
@@ -300,8 +300,8 @@ mixin _BootstrapMixin on _CleanMixin {
             collectionStyle: CollectionStyle.BLOCK,
           ),
         );
+        didUpdate = true;
       }
-      didUpdate = true;
     }
 
     return didUpdate;
--- a/packages/melos/lib/src/commands/bootstrap.dart
+++ b/packages/melos/lib/src/commands/bootstrap.dart
@@ -35,6 +35,10 @@ mixin _BootstrapMixin on _CleanMixin {
         }
 
         try {
+          if (workspace.config.commands.bootstrap.shareDependencies) {
+            await _setSharedDependencies(workspace);
+          }
+
           await _linkPackagesWithPubspecOverrides(workspace);
         } on BootstrapException catch (exception) {
           _logBootstrapException(exception, workspace);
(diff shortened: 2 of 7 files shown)


lohnn:
Another question: should Melos maybe sync the Dart and Flutter version in environment as well?

vkammerer:
My two cents:
- Instead of a dedicated separate file, we could include these properties in `melos.yaml`. It is a feature of melos, so there is no reason to exclude it from the melos settings. At a later point, this could potentially also be a part of `pubspec.yaml` if https://github.com/invertase/melos/issues/472 is implemented.
- the name `shareDependencies` makes sense if the dependencies versions are taken from the root package's own dependencies, as suggested by @blaugold in https://github.com/invertase/melos/issues/94#issuecomment-1539961842, but since your implementation follows another approach, there is no "sharing" anymore. Instead, a more appropriate name could be `assignCommonDependencies` or similar.
- you could avoid calling https://github.com/invertase/melos/pull/526/files#diff-b5438310509b0fb6f04ef9776c2b9d3ec2a18fe0e0bc7165cdda954148ad9a99R240-R243 if `dependenciesUpdated == 0 && devDependenciesUpdated == 0`, (which will be the case most of the time, since `melos bootstrap` is likely to be run more often than common dependencies version are updated).
- There is a mixed usage of the terms "package" and "dependency", which is confusing. I may be wrong, but it looks to me that a "package" refers to a local package managed by melos, while a "dependency" refers to an entry of the "dependencies" and "dev_dependencies" fields. If that is indeed the case, there are several usages of "package" that should be replaced with "dependency":
  - if we wanted to go with a separate file, I guess it should be named "common_dependencies.yaml" rather than "common_packages.yaml"
  - `_updatePackages` -> `_updateDependencies`
  - `packagesToUpgrade` -> `dependenciesToUpgrade`
  - etc
- the naming of `dependenciesUpdated` and `devDependenciesUpdated` sound like booleans to me. I would suggest `updatedDependenciesCount` and `updatedDevDependenciesCount` to be more explicit.

But all these are details, compared to the more important design decision of whether we want to share the root package's dependencies or use separate fiels as you have done. As I have already expressed in https://github.com/invertase/melos/issues/94#issuecomment-1543268061 , I am more in favour of new fields, but it would be nice to hear @blaugold 's opinion on this design choice.

lohnn:
I would also love to hear what @blaugold has to say as I should have some time over to continue with this PR next week.

blaugold:
Sorry for not responding earlier! 😔 The past few weeks have been really busy.

@lohnn Thanks for working on this!

> Another question: should Melos maybe sync the Dart and Flutter version in environment as well?

Great idea. I suspect it's just a question of time until someone asks for this, especially with support for common dependencies.

@vkammerer Thanks for the detailed input. You convinced me that putting the common dependencies into the root pubspec is probably not a good idea. I feel like the next best option is to put them into `melos.yaml`. Like you, I don't really see a reason not to. We already maintain a JSON schema for `melos.yaml`, which can be used for autocompletion and validation. Extending that is much easier than introducing a new type of file.

Regarding naming, I would be fine with mirroring that of `pubspec.yaml` under the bootstrap command options:

```yaml
command:
  bootstrap:
    environment:
      ...
    dependencies:
      ...
    devDependencies:
      ...
```

lohnn:
Heh, it's all good! I'm getting married soon and know exactly what you mean. Thankfully I can do some of this job during work hours (wish I could say the same for my own package, but we currently don't use it at work 😅)

> Great idea. I suspect it's just a question of time until someone asks for this, especially with support for common dependencies.

Then I'll add it to the TODO list.

Regarding putting the common dependencies in `melos.yaml` file I only have one gripe. The `melos.yaml` file is, at least in our project, already quite bloated clocking in at just short of 200 ilnes. Adding all the common dependencies would increase the size of this file even more.

On the flip side, we all use IDEs and can collapse parts of the file if we want to make it more readable, so if you still wish for the common dependencies (and version number) to live in the `melos.yaml` file, I'll update this PR accordingly :)

> Regarding naming, I would be fine with mirroring that of pubspec.yaml under the bootstrap command options:

I imagine you mean it should be:

```yaml
command:
  bootstrap:
    environment:
      ...
    dependencies: 
      ...
    dev_dependencies: # instead of devDependencies
      ...
```

vkammerer:
@lohnn I would much prefer the simplicity of a few lines added to `melos.yaml` (which would be visible only when you view that file) than having to deal with yet another config file just for this purpose (which would always be visible when you browse your project in your code editor, which is to say almost all the time).

Moreover, the argument about it belonging to melos responsibilities and therefore belonging to `melos.yaml` still stands.

lohnn:
Check! I'll take it and update the code accordingly 👍 (whenever I too get time 🤦‍♂️)

lohnn:
And regarding tests for this work. I should look into that I guess?

lohnn:
I have now added a test to validate that the bootstrap dependency sync runs as intended, and the PR is now ready to be reviewed.

lohnn:
Ping: I would love for this PR to be reviewed.

blaugold:
@lohnn Sorry for the delay! I'll review it tomorrow.

packages/melos/lib/src/commands/bootstrap.dart
```diff
@@ -199,6 +214,127 @@ mixin _BootstrapMixin on _CleanMixin {
     }
   }
 
+  Future<void> _setSharedPreferencesInPackage(
```
blaugold, line 217: Consider the name `_setSharedDependenciesForPackage`. `Environment` also specifies dependencies, just on SDKs not packages.

packages/melos/lib/src/commands/bootstrap.dart
```diff
@@ -199,6 +214,127 @@ mixin _BootstrapMixin on _CleanMixin {
...
+    );
+
+    final updatedDevDependenciesCount = _updateDependencies(
+      pubspecEditor: pubspecEditor,
+      workspaceDependencies: devDependencies,
+      packageDependencies: package.pubSpec.devDependencies,
+      pubspecKey: 'dev_dependencies',
+    );
+
+    await writeTextFileAsync(
```
blaugold, line 247: Can we avoid writing `pubspec.yaml` when we did not update anything?

packages/melos/lib/src/commands/bootstrap.dart
```diff
@@ -199,6 +214,127 @@ mixin _BootstrapMixin on _CleanMixin {
...
+    final message = <String>[
+      if (updatedEnvironment) 'Updated environment',
+      if (updatedDependenciesCount > 0)
+        'Updated $updatedDependenciesCount dependencies',
+      if (updatedDevDependenciesCount > 0)
+        'Updated $updatedDevDependenciesCount dev_dependencies',
+    ];
+    if (message.isNotEmpty) {
+      logger.child(packageNameStyle(package.name), prefix: '').child(
+            message.join('\n'),
```
blaugold, line 261: Removing the trailing comma in this line makes the whole log statement format a little nicer.

packages/melos/lib/src/commands/bootstrap.dart
```diff
@@ -199,6 +214,127 @@ mixin _BootstrapMixin on _CleanMixin {
...
+
+  int _updateDependencies({
+    required YamlEditor pubspecEditor,
+    required Map<String, DependencyReference>? workspaceDependencies,
+    required Map<String, DependencyReference> packageDependencies,
+    required String pubspecKey,
+  }) {
+    if (workspaceDependencies == null) return 0;
+    // Filter out the packages that does not exist in package and only the
+    // dependencies that has a different version specified in the workspace.
```
blaugold, line 318: does => do

has => have

packages/melos/lib/src/common/utils.dart
```diff
@@ -321,6 +321,9 @@ String melosOverridesYamlPathForDirectory(String directory) =>
 String pubspecPathForDirectory(String directory) =>
     p.join(directory, 'pubspec.yaml');
 
+String pubspecCommonPathForDirectory(String directory) =>
+    p.join(directory, 'common_packages.yaml');
+
```
blaugold, line 326: This is not used anywhere.

docs/commands/bootstrap.mdx
```diff
@@ -84,6 +84,45 @@ example:
...
+dependencies:
+  collection: ^1.17.0
+  uni_links2:
+  uni_links_macos:
+    git: https://github.com/SamJakob/uni_links_macos.git
+
+dev_dependencies:
+  build_runner: ^2.3.3
+# ...
+```
```
blaugold, line 124: This looks outdated since it talks about `common_packages.yaml`.

blaugold (approved):
LGTM

Thanks! ❤️

feinstein:
I am trying to find this mechanism in the docs website, but there I can only see a `dependencies` section in `bootstrap`. Has the `common_dependencies.yaml` file being removed?

PR #618, opened 2023-12-21
chore(melos): updated dependencies to latest stable versions

OutdatedGuy:
## Description

Updated packages [file](https://pub.dev/packages/file) and [pub_updater](https://pub.dev/packages/pub_updater) to latest stable versions i.e. `v7.0.0` and `v0.4.0` respectively.

Also updated dependency versions in the `melos.yaml` file to properly bootstrap melos while using `melos bootstrap`.

## Type of Change

- [x] 🗑️ `chore` -- Chore

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -17,7 +17,7 @@ command:
       cli_util: '>=0.3.0 <0.5.0'
       collection: ^1.14.12
       conventional_commit: ^0.6.0+1
-      file: ^6.1.0
+      file: ^7.0.0
       glob: ^2.1.0
       graphs: ^2.1.0
       http: ">=0.13.1 <2.0.0"
@@ -28,7 +28,7 @@ command:
       pool: ^1.4.0
       prompts: ^2.0.0
       pub_semver: ^2.0.0
-      pub_updater: ^0.3.0
+      pub_updater: ^0.4.0
       pubspec: ^2.1.0
       string_scanner: ^1.0.5
       yaml: ^3.1.0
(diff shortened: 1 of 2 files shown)


spydon (changes requested):
This does not pass the version constraints.

spydon (commented):
I don't think this change gives us anything, it will be more work maintaining those dependency ranges. It's better to wait with bumping these dependencies until we bump Dart.

OutdatedGuy:
Yo @spydon, this is the same tactic we used while updating the `http` package in [this commit](https://github.com/invertase/melos/commit/979d0b157b4cffa4d91882be3a3a5be03dcbd147).

Also because of these constraints, I'm unable to add melos as I'm using packages with the latest versions of `pub_updater` and `file`.

Additionally, we can later remove the range after the current melos is updated.

spydon:
> Yo @spydon, this is the same tactic we used while updating the `http` package in [this commit](https://github.com/invertase/melos/commit/979d0b157b4cffa4d91882be3a3a5be03dcbd147).
> 
> Also because of these constraints, I'm unable to add melos as I'm using packages with the latest versions of `pub_updater` and `file`.
> 
> Additionally, we can later remove the range after the current melos is updated.

The dependencies of Melos should not affect your usage of it? If you have a package/app in the root I recommend that you move it into a sub-directory, for example `packages/` for packages or `apps/` for apps, because you shouldn't share the Melos workspace pubspec.

PR #648, opened 2024-02-23
refactor: move environment variable related strings into one class

Pavel-Sulimau:
## Description

Extracted the duplicated and spread through different places environment keys into one class to have a more maintainable approach.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/common/environment_variable_key.dart
+++ b/packages/melos/lib/src/common/environment_variable_key.dart
@@ -1,0 +1,44 @@
+class EnvironmentVariableKey {
+  EnvironmentVariableKey._();
+
+  static const String melosRootPath = 'MELOS_ROOT_PATH';
+  static const String melosPackageName = 'MELOS_PACKAGE_NAME';
+  static const String melosPackageVersion = 'MELOS_PACKAGE_VERSION';
+  static const String melosPackagePath = 'MELOS_PACKAGE_PATH';
+  static const String melosParentPackageName = 'MELOS_PARENT_PACKAGE_NAME';
+  static const String melosParentPackageVersion =
+      'MELOS_PARENT_PACKAGE_VERSION';
+  static const String melosParentPackagePath = 'MELOS_PARENT_PACKAGE_PATH';
+  static const String melosScript = 'MELOS_SCRIPT';
+  static const String melosTest = 'MELOS_TEST';
+
+  /// This user-defined environment variable contains a comma delimited list of
+  /// package names that Melos should focus on. This will act as the global
+  /// `scope` package filter, and it will override the `scope` for all the
+  /// filtering options defined in the `packageFilters` section.
+  static const String melosPackages = 'MELOS_PACKAGES';
+
+  /// This user-defined environment has a path to the Dart/Flutter SDK to use.
+  /// This environment variable has precedence over the `sdkPath` option in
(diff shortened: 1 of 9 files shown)


packages/melos/lib/src/common/environment_variable_key.dart
```diff
@@ -0,0 +1,44 @@
+class EnvironmentVariableKey {
```
spydon, line 1: Wouldn't this be better as an enhanced enum?

spydon (commented):
Nice refactor! Just one comment

packages/melos/lib/src/common/environment_variable_key.dart
Pavel-Sulimau: @spydon, initially, I did think of using an Enum for it, i.e.,
```dart
enum EnvironmentVariableKey {
  melosRootPath('MELOS_ROOT_PATH'),
  melosPackageName('MELOS_PACKAGE_NAME'),
  melosPackageVersion('MELOS_PACKAGE_VERSION');
  // the rest of the values...

  const EnvironmentVariableKey(this._value);

  final String _value;

  @override
  String toString() => _value;
}
```
but decided not to do so because of these 2 reasons:
1. A `String` is expected in every place the env values are used at the moment and with such an enum it is possible to access not only the value, i.e. `EnvironmentVariableKey.melosRootPath.toString()` or `EnvironmentVariableKey.melosRootPath.value`, if `value` made public, but also accidentally use a name of the value, i.e. `EnvironmentVariableKey.melosRootPath.name`. It seemed to me like a more error-prone approach for the current case.
2. Since strings are expected in all the cases, I'd have to either use a `toString` on the enum values or make the `value` public which results in longer lines compared to the approach with static const strings.
`EnvironmentVariableKey.melosRootPath` vs `EnvironmentVariableKey.melosRootPath.value` / `EnvironmentVariableKey.melosRootPath.toString()`

Does it sound reasonable?

spydon: I think strings should be used as little as possible for these cases, isn't it possible to just have a `fromString` method in the enum and use it everywhere there are comparisons instead? Then there isn't any need for string comparisons, and it looks like you've already removed most of the string comparisons anyways?

Pavel-Sulimau: >  it looks like you've already removed most of the string comparisons anyways?

Not really, the env variable values are used as strings everywhere at the moment: in maps, i.e. `Map<String, String>`, in string interpolations, i.e. '%${EnvironmentVariableKey.melosScript}%'`, so @spydon, I do not see how we can easily use the Enum everywhere, at least without a way bigger refactoring.

@Salakar, would you like to share your opinion as well?

spydon: Ah alright, let's merge this first then and then turn it into an enum then, this is one step in the right direction at least!

spydon (approved):

packages/melos/lib/src/common/environment_variable_key.dart
Pavel-Sulimau: Yep, agree, this is indeed a step to a more maintainable code base, let's get it merged then!

PR #655, opened 2024-03-10
feat: built-in melos command for analyzing projects

### LINKED ISSUE #424: request:  `melos analyze`

### Command

```
melos analyze [--fatal-infos] [--[no-]fatal-warnings]
``` 

### Description

It would be nice to have a built-in Melos command for analyzing projects. The command should accept `--fatal-infos` and `--[no-]fatal-warnings` arguments.

### Reasoning

I think it would be useful because I have such a script for every project of mine using Melos. :wink:

It's common for Melos projects to define a script that runs either `dart analyze` or `flutter analyze`, something like:
```yaml
analyze: melos exec -c1 -- flutter analyze --fatal-infos .
```

jessicatarra:
## Description

First things first, I'm really excited to work on this issue because this year I made a personal goal to contribute more to open source projects. It makes me happy to be able to do it for a tool that I use every day. 

Usually, melos projects use a script to run either `dart analyze` or `flutter analyze`, as mentioned by @jpnurmi  in issue #424. However, I agree that it would be better to have a built-in command for this. Some newbies to the tool might try running this command and then realize they need extra setup to get the same results.

So, this pull request aims to address this problem by adding the needed functionality. It also includes features like running tasks concurrently, filtering, and support for `--fatal-infos` and `--[no-]fatal-warnings` options.

I also took the time to update the readme file and provide the latest help section, as well as update the `melos.yaml` file and remove the old way to run the analyzer.

Dear maintainers, please feel free to review, modify, or add code to this contribution :)

## Type of Change
- [ ] 🗑️ `chore` -- Chore

Closes #424

### DIFF
--- a/packages/melos/lib/src/command_runner/analyze.dart
+++ b/packages/melos/lib/src/command_runner/analyze.dart
@@ -28,8 +28,8 @@ class AnalyzeCommand extends MelosCommand {
     argParser.addFlag(
       'fatal-infos',
       negatable: false,
-      help: 'Enables treating info-lever issues as fatal errors,'
-          ' stopping the process if any are encountered.',
+      help: 'Enables treating info-lever issues as fatal errors, '
+          'stopping the process if any are encountered.',
     );
 
     argParser.addFlag(
--- a/packages/melos/lib/src/command_runner/analyze.dart
+++ b/packages/melos/lib/src/command_runner/analyze.dart
@@ -15,8 +15,6 @@
  *
  */
 
-import 'dart:io';
-
 import '../commands/runner.dart';
 import '../common/utils.dart';
 import 'base.dart';
@@ -50,15 +48,6 @@ class AnalyzeCommand extends MelosCommand {
 
   @override
   Future<void> run() async {
-    if (argResults!.rest.isNotEmpty) {
-      logger.log('No additional command line arguments are needed');
-      logger.log(description);
-      logger.log(argParser.usage);
-
-      exitCode = 1;
-      return;
-    }
-
     final fatalInfos = argResults?['fatal-infos'] as bool;
     final fatalWarnings = argResults!.optional('fatal-warnings') as bool?;
     final concurrency = int.parse(argResults!['concurrency'] as String);
--- a/packages/melos/lib/src/command_runner/analyze.dart
+++ b/packages/melos/lib/src/command_runner/analyze.dart
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2016-present Invertase Limited & Contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this library except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import 'dart:io';
+
+import '../commands/runner.dart';
+import '../common/utils.dart';
+import 'base.dart';
+
+class AnalyzeCommand extends MelosCommand {
+  AnalyzeCommand(super.config) {
+    setupPackageFilterParser();
+    argParser.addOption('concurrency', defaultsTo: '1', abbr: 'c');
+    argParser.addFlag(
+      'fatal-infos',
+      negatable: false,
+      help: 'Enables treating info-lever issues as fatal errors,'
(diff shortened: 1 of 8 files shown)


packages/melos/lib/src/command_runner/analyze.dart
```diff
@@ -0,0 +1,74 @@
...
+
+class AnalyzeCommand extends MelosCommand {
+  AnalyzeCommand(super.config) {
+    setupPackageFilterParser();
+    argParser.addOption('concurrency', defaultsTo: '1', abbr: 'c');
+    argParser.addFlag(
+      'fatal-infos',
+      negatable: false,
+      help: 'Enables treating info-lever issues as fatal errors,'
+          ' stopping the process if any are encountered.',
```
spydon, line 32:
```suggestion
      help: 'Enables treating info-lever issues as fatal errors, '
          'stopping the process if any are encountered.',
```

packages/melos/lib/src/command_runner/analyze.dart
```diff
@@ -0,0 +1,74 @@
...
+          'When enabled, any warning will cause the command to fail.',
+    );
+  }
+
+  @override
+  final String name = 'analyze';
+
+  @override
+  final String description =
+      '''Analyzes all packages in your project for potential issues in a single run. Optionally configure severity levels. Supports all package filtering options.''';
```
spydon, line 47: Split this on multiple lines

packages/melos/lib/src/command_runner/analyze.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  @override
+  Future<void> run() async {
+    if (argResults!.rest.isNotEmpty) {
+      logger.log('No additional command line arguments are needed');
```
spydon, line 52: This might be a bit confusing if one of the valid arguments are used and then an invalid one is also used, maybe list the invalid ones?

spydon (commented):
Very nice! Could you add some documentation in a new `.mdx` file too, similarly to the documentation for the other commands?

packages/melos/lib/src/command_runner/analyze.dart
jessicatarra: I think it's pretty hard to come up with a list of invalid arguments because it can be anything, right? That's why I chose to remove this code. Even if the user tries to pass an extra argument, it simply won't work, and the 'melos analyze' command will run with the 'default' behavior, which means it will analyze every package or accept a valid argument. For example ⬇️ 
Screenshot 2024-03-10 at 20 50 16

Also, it's important to mention that this made me wonder what would happen if the user combines two arguments, like `--fatal-infos` and `--fatal-warnings`. So, I introduced the following commit: e66abd61e1b9b2e41cb28b9fd0358989bde6936d

spydon: Since it has the list of user arguments it could just print the invalid ones for the user, but I think it was better to remove it like you did anyways, then it's more in line with how the other commands look like.

jessicatarra:
Yes sure, I just added the proper documentation for the analyzer command @spydon. Commit: 963b7022a1cde055da319fe8c71370ad89ec3b55

spydon:
I wonder if we might need to treat this as a breaking change, even though it technically isn't one, because so many of our users have a script called analyze. What happens if the user has a script called analyze already?

jessicatarra:
It's a good question, so I did some tests, and it seems that running just `melos analyze` will trigger the built-in command, while running `melos run analyze` will execute the script. I'm also aware that you can call a script without using the `run` keyword. However, if the user is specifically using the version with the analyzer command, it will fall into the first situation I mentioned earlier. What we can do, and this is my suggestion, is to let the user know that there's a better way to trigger the analyzer when they attempt to use it through the script, what do you think?

spydon:
> and this is my suggestion, is to let the user know that there's a better way to trigger the analyzer when they attempt to use it through the script, what do you think?

Yeah, that sounds good. Most users will probably not even notice that would be using the built-in command instead of their own after they have updated, until they read about it. Since we don't fail fatally we'll hopefully not crash everyone's pipelines, so I think we're fine. :sweat_smile:

spydon (approved):

Salakar (approved):

Salakar:
Very nice! Thanks @jessicatarra for doing this 🎉

renefloor:
> Most users will probably not even notice that would be using the built-in command instead of their own after they have updated, until they read about it.

@spydon I would have preferred melos to block the analyze command if it's duplicated in the melos.yaml. Now our CI started to fail and it was not immediately clear why. It's just silently overwriting our own command and not giving any warning about that.

spydon:
> @spydon I would have preferred melos to block the analyze command if it's duplicated in the melos.yaml. Now our CI started to fail and it was not immediately clear why. It's just silently overwriting our own command and not giving any warning about that.

That's why we did a major version bump, since you then have to actively go into your root pubspec and bump Melos to the next major version. But you can open a feature request for allowing `analyze` and `format` to be overwritten.

PR #664, opened 2024-03-17
feat: add support for combining scripts

### LINKED ISSUE #653: Combining scripts

### Command

_No response_

### Description

Currently, if you want one script to execute a set of other scripts, you can do something like:
```yaml
  test:
    description: Run all tests.
    run: >-
      melos run test:unit &&
      melos run test:widget &&
      melos run test:lints &&
      melos run test:integration
```
However, since these are simply invoked in a subprocess (without a tty) the color coding you get when invoking the scripts separately is lost.

Also, only `&&` is really safe to use, so the construct is implicitly serialised, and fail-fast in nature.

I suggest some way of running a set of other scripts, allowing `melos`

- to avoid calling itself as subprocess, 
- be smarter about whether to fail or proceed, 
- and not loose the tty (color, etc.)

As above, this would also mitigate some use of multiline scripts, which come with its own set of problems.

### Reasoning

I find the `melos run x && melos run y` useful, but cumbersome, and ruining the nice output. It is definitely a nice-to-have.

_No response_

jessicatarra:
## Description

This PR adds the capability to combine multiple scripts into a single script definition. In the provided example, the `pre-commit` script is configured to invoke a simple command as `echo 'hello world'`, and also individual scripts such as `format` and `analyze`.

### YAML file
```yaml
scripts:
  pre-commit:
    description: pre-commit git hook script
    steps:
      - echo 'hello world'
      - format
      - analyze
  format:
    description: Format Dart code.
    run: dart format .
    exec:
      concurrency: 5
  analyze:
    description: Analyze Dart code
    run: dart analyze .
    exec:
      concurrency: 5
```

### Output
```console
┌─(~/Development/melos)──────────────────────────────────────────────────────────────────────────────────────────────────────────(jessica@jessicas-MacBook-Air:s005)─┐
└─(21:14:01 on feature/multiple-steps-script ✹)──> melos run pre-commit                                                                                ──(Sat,Mar16)─┘
Building package executable... 
Built melos:melos.
melos run pre-commit
  └> echo 'hello world'
     └> RUNNING

hello world

melos run pre-commit
  └> echo 'hello world'
     └> SUCCESS

melos run pre-commit
  └> format
     └> RUNNING

melos run format
  └> melos exec --concurrency 5 -- "dart format ."
     └> RUNNING

$ melos exec
  └> dart format .
     └> RUNNING (in 3 packages)

-----------------------------------------------------------------------------------------------------------------------------------------------------------------------
[conventional_commit]: Formatted 3 files (0 changed) in 0.18 seconds.
[melos]: Formatted 85 files (0 changed) in 0.60 seconds.
[melos_workspace]: Formatted 90 files (0 changed) in 0.61 seconds.
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------

$ melos exec
  └> dart format .
     └> SUCCESS

melos run format
  └> melos exec --concurrency 5 -- "dart format ."
     └> SUCCESS

melos run pre-commit
  └> format
     └> SUCCESS

melos run pre-commit
  └> analyze
     └> RUNNING

melos run analyze
  └> melos exec --concurrency 5 -- "dart analyze ."
     └> RUNNING

$ melos exec
  └> dart analyze .
     └> RUNNING (in 3 packages)

-----------------------------------------------------------------------------------------------------------------------------------------------------------------------
[conventional_commit]: Analyzing ....
[melos]: Analyzing ....
[melos_workspace]: Analyzing ....
[conventional_commit]: No issues found!
[melos]: No issues found!
[melos_workspace]: No issues found!
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------

$ melos exec
  └> dart analyze .
     └> SUCCESS

melos run analyze
  └> melos exec --concurrency 5 -- "dart analyze ."
     └> SUCCESS

melos run pre-commit
  └> analyze
     └> SUCCESS
```
As it stands, when using the multiple script definition, we are not allowed to use execution options like concurrency or `packageFilter` option, because that would increase the level of complexity and could cause several problems, so my suggestion for now is to be able to configure these things in the individual scripts.

Closes #653 

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -21,11 +21,6 @@ mixin _RunMixin on _Melos {
     }
 
     if (script.steps != null && script.steps!.isNotEmpty) {
-      //TODO(jessica): Reevaluate the design decision regarding execution
-      // options (e.g., concurrency) for scripts with steps.
-      // Current concern: Allowing execution options for scripts with steps
-      // might introduce complexity and unpredictability, especially if nested
-      // scripts also utilize concurrency.
       if (script.exec != null) {
         throw ScriptExecOptionsException._(
           scriptName,
@@ -82,7 +77,6 @@ mixin _RunMixin on _Melos {
   /// eventually leads back to the original script, it indicates a
   /// recursive script call, which can result in an infinite loop during
   /// execution.
-  ///
   void _detectRecursiveScriptCalls(Script script) {
     final visitedScripts = <String>{};
 
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -31,6 +31,9 @@ mixin _RunMixin on _Melos {
           scriptName,
         );
       }
+
+      _detectRecursiveScriptCalls(script);
+
       await _runMultipleScripts(
         script,
         global: global,
@@ -70,6 +73,37 @@ mixin _RunMixin on _Melos {
       throw ScriptException._(script.name);
     }
     resultLogger.child(successLabel);
+  }
+
+  /// Detects recursive script calls within the provided [script].
+  ///
+  /// This method recursively traverses the steps of the script to check
+  /// for any recursive calls. If a step calls another script that
+  /// eventually leads back to the original script, it indicates a
+  /// recursive script call, which can result in an infinite loop during
+  /// execution.
+  ///
+  void _detectRecursiveScriptCalls(Script script) {
+    final visitedScripts = <String>{};
+
+    void traverseSteps(Script currentScript) {
+      visitedScripts.add(currentScript.name);
+
+      for (final step in currentScript.steps!) {
+        if (visitedScripts.contains(step)) {
+          throw RecursiveScriptCallException._(step);
+        }
+
+        final nestedScript = config.scripts[step];
+        if (nestedScript != null) {
+          traverseSteps(nestedScript);
+        }
+      }
+
+      visitedScripts.remove(currentScript.name);
+    }
+
+    traverseSteps(script);
   }
 
   Future<String> _pickScript(MelosWorkspaceConfig config) async {
@@ -220,6 +254,8 @@ mixin _RunMixin on _Melos {
     for (final step in steps) {
       final scriptCommand =
           scripts.containsKey(step) ? 'melos run $step' : step;
+
+      /// TODO: return scripts.containsKey(step) if true
       final scriptSourceCode = targetStyle(
         step.withoutTrailing('\n'),
       );
@@ -341,5 +377,18 @@ class MissingScriptCommandException implements MelosException {
         'This can be done by filling "run" with a command, '
         'defining a sequence of commands in the "steps", '
         'or by providing a script execution definition in the "exec".';
+  }
+}
+
+class RecursiveScriptCallException implements MelosException {
+  RecursiveScriptCallException._(this.scriptName);
+
+  final String scriptName;
+
+  @override
+  String toString() {
+    return 'RecursiveScriptCallException: Detected a recursive call in script '
+        'execution. The script "$scriptName" calls itself or forms a recursive '
+        'loop.';
   }
 }
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -41,7 +41,7 @@ mixin _RunMixin on _Melos {
       return;
     }
 
-    if (script.run!.isEmpty && script.exec is! String) {
+    if (script.run == null && script.exec is! String) {
       throw MissingScriptCommandException._(
         scriptName,
       );
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -255,7 +255,6 @@ mixin _RunMixin on _Melos {
       final scriptCommand =
           scripts.containsKey(step) ? 'melos run $step' : step;
 
-      /// TODO: return scripts.containsKey(step) if true
       final scriptSourceCode = targetStyle(
         step.withoutTrailing('\n'),
       );
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -20,6 +20,33 @@ mixin _RunMixin on _Melos {
       );
     }
 
+    if (script.steps != null && script.steps!.isNotEmpty) {
+      //TODO(jessica): Reevaluate the design decision regarding execution
+      // options (e.g., concurrency) for scripts with steps.
+      // Current concern: Allowing execution options for scripts with steps
+      // might introduce complexity and unpredictability, especially if nested
+      // scripts also utilize concurrency.
+      if (script.exec != null) {
+        throw ScriptExecOptionsException._(
+          scriptName,
+        );
+      }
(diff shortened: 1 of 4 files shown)


spydon:
> As it stands, when using the multiple script definition, we are not allowed to use execution options like concurrency or packageFilter option, because that would increase the level of complexity and could cause several problems, so my suggestion for now is to be able to configure these things in the individual scripts.

Completely agree!

packages/melos/lib/src/scripts.dart
```diff
@@ -129,15 +130,18 @@ class Script {
     required String workspacePath,
   }) {
     final scriptPath = 'scripts/$name';
-    String run;
+    var run = '';
     String? description;
     var env = <String, String>{};
+    var steps = <String>[];
```
spydon, line 136: This could be final, and the steps could just be added in the `if` further down.
You probably just follow how `env` is done, but that one should probably also change to be final (not in necessarily baked into this PR though).

packages/melos/lib/src/scripts.dart
```diff
@@ -129,15 +130,18 @@ class Script {
     required String workspacePath,
   }) {
     final scriptPath = 'scripts/$name';
-    String run;
+    var run = '';
```
spydon, line 133: I would define this where the `if` is and have it as final too, and set the default value in the else case.

spydon: What do you think about having it as nullable instead? So that it works in the same way as the `description`?

packages/melos/test/commands/run_test.dart
```diff
@@ -361,5 +361,195 @@ melos run test_script
...
+      );
+
+      expect(() => melos.run(scriptName: 'test_script'), throwsException);
+    });
+  });
+
+  group('multiple scripts', () {
+    test(
+        'verifies that a melos script can successfully call another '
+        'script as a step and execute commands', () async {
```
spydon, line 437: Could you add a test where another script that contains `steps` is included in the list of `steps`?

spydon (commented):
Great job! Only some minor comments.

Should we maybe have a simple check that the steps list doesn't contain the script itself, so that the user doesn't get stuck in an infinite loop? Maybe it would be worth checking if it ends up in an infinite loop if the start script is a later step in a nested tree too, but that could be done later.

jessicatarra:
Hey @spydon, thanks for your feedback! and regarding the test scenario you just mentioned in the first comment, I think it's worth including a validation to avoid an infinite loop, because that's actually what ends up happening, so I'm working on it.

packages/melos/lib/src/scripts.dart
jessicatarra: Yes agree! I make it nullable instead as you mentioned in commit 5c19f3da0d50534a64ac035ef2e85e601e2a43a9

packages/melos/lib/src/scripts.dart
jessicatarra: if `steps` become an empty final variable declaration and then added further down in the `if` statement, it will actually cause the following error:

```
The final variable 'steps' can't be read because it's potentially unassigned at this point. (Documentation)  Ensure that it is assigned on necessary execution paths.
```

I tested some approaches, like the following, for example:
```
if (yaml is String) {
  run = yaml;
}

if (yaml is! Map<Object?, Object?>) {
  throw MelosConfigException('Unsupported value for script $name');
}

[...]
```
This way we will reverse the logic and expected positive result at the end without if statements, and a final variable for `steps` or `env` will work. But I'm missing something because in theoric should work but it's breaking everything ahahah

packages/melos/test/commands/run_test.dart
jessicatarra: I added two tests: one tests the scenario where a script with steps calls another script with steps, and the other test expects an exception related to detecting a recursive call in the script that leads to an infinite loop 17b701a2e692e8e8664b9cfa8f596da0bd5ff0f1.

packages/melos/lib/src/scripts.dart
spydon: It would need an `else` case too, then the compiler will know that it can be final and assigned.

spydon: > But I'm missing something because in theoric should work but it's breaking everything ahahah

Hahah classic! I'm at my phone at the moment, but I can have a look tomorrow if you haven't solved it, I bet you will though. 😄

jessicatarra: Hey @spydon, I was able to figured it out! I like the first approach I just shared with you because we're going to get rid of the nested if statements and apply a sort of 'early return pattern', which was also a big help for me in order to  understand what the `Script.fromYaml` factory was trying to achieve.

In the commit e5f44b20939d99910c8815d0691ef398fc433ad3, I introduce this new approach and additionally, I made the `steps` variable a `final` one since it is assigned a value and doesn't change afterward. Let me know what you think.

packages/melos/lib/src/commands/run.dart
```diff
@@ -20,6 +20,36 @@ mixin _RunMixin on _Melos {
       );
     }
 
+    if (script.steps != null && script.steps!.isNotEmpty) {
+      //TODO(jessica): Reevaluate the design decision regarding execution
+      // options (e.g., concurrency) for scripts with steps.
+      // Current concern: Allowing execution options for scripts with steps
+      // might introduce complexity and unpredictability, especially if nested
+      // scripts also utilize concurrency.
```
spydon, line 28: This TODO can be removed, I don't think we'd ever want to enable concurrency for groups, it'll be way too messy. :sweat_smile:

packages/melos/lib/src/commands/run.dart
```diff
@@ -45,6 +75,37 @@ mixin _RunMixin on _Melos {
...
   }
 
+  /// Detects recursive script calls within the provided [script].
+  ///
+  /// This method recursively traverses the steps of the script to check
+  /// for any recursive calls. If a step calls another script that
+  /// eventually leads back to the original script, it indicates a
+  /// recursive script call, which can result in an infinite loop during
+  /// execution.
+  ///
```
spydon, line 85:
```suggestion
```

spydon (approved):
Lgtm, just two very minor comments.
Really looking forward to using this feature!

spydon:
> Isn't this already possible like:
> 
> ```
> pre_commit:
>  description: pre-commit git hook script
>     run: |
>       melos run format
>       melos run analyze
>       melos run test
>   exec:
>     concurrency: 3
>     failFast: true
> ```

Sure you can do that, but this is a much cleaner API!
Especially since multi-line scripts can have some quirks, which even your example suffers from (missing `&&`). :smile:

Salakar (approved):
Woah this is nice, thanks for all the contributions you're shipping to Melos! 🚀

provokateurin:
Cool feature, can we also have this for hooks? I guess this would be a breaking change so I'd understand if you wouldn't want to add it, but I think it would be very useful.

spydon:
> Cool feature, can we also have this for hooks? I guess this would be a breaking change so I'd understand if you wouldn't want to add it, but I think it would be very useful.

You can open an enhancement issue regarding that if you want, so that we don't have the discussion about it on this PR.
For now, you could just create a melos script with `steps` and call it from the hook instead.

PR #683, opened 2024-03-29
feat: support for melos command within script steps

jessicatarra:
## Description

This pull request introduces support for melos command within script steps, as suggested in issue #653. An example configuration demonstrating this new capability is as follows:

```yaml
scripts:
  pre-commit:
    description: pre-commit git hook script
    steps:
      - echo 'hello world'
      - format --output none --set-exit-if-changed
      - analyze --fatal-infos
``` 

I'll need some help on determining the appropriate range of commands to support. Should we use every command possible?  
Currently, the following commands are under consideration:

```dart
    const melosCommands = {
      'analyze',
      'format',
      'bs',
      'bootstrap',
      'clean',
      'list',
      'publish',
    };

```

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -313,7 +313,6 @@ mixin _RunMixin on _Melos {
 
     if (exitCode != 0) {
       resultLogger.child(failedLabel);
-      throw ScriptException._(script.name);
     } else {
       resultLogger.child(successLabel);
     }
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -258,12 +258,12 @@ mixin _RunMixin on _Melos {
   }
 
   String _buildScriptCommand(String step, Scripts scripts) {
-    if (_isStepACommand(step)) {
-      return 'melos $step';
-    }
-
     if (scripts.containsKey(step)) {
       return 'melos run $step';
+    }
+
+    if (_isStepACommand(step)) {
+      return 'melos $step';
     }
 
     return step;
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -239,6 +239,33 @@ mixin _RunMixin on _Melos {
     await _executeScriptSteps(steps, scripts, script, environment);
   }
 
+  /// Checks if the given [step] is a recognized Melos command.
+  bool _isStepACommand(String step) {
+    const melosCommands = {
+      'analyze',
+      'format',
+      'bs',
+      'bootstrap',
+      'clean',
+      'list',
+      'publish',
+    };
+
+    return melosCommands.contains(step);
+  }
+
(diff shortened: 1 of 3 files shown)


packages/melos/lib/src/commands/run.dart
```diff
@@ -239,15 +239,44 @@ mixin _RunMixin on _Melos {
...
+  }
+
+  String _buildScriptCommand(String step, Scripts scripts) {
+    if (_isStepACommand(step)) {
+      return 'melos $step';
+    }
+
+    if (scripts.containsKey(step)) {
+      return 'melos run $step';
+    }
```
spydon, line 267: Should these ifs be in the opposite order maybe? So if the user has accidentally defined a script with the same name as a command, it would still work?

spydon (approved):
Lgtm!

packages/melos/lib/src/commands/run.dart
jessicatarra: I introduced this suggestion in commit 0ca796941c449500bdbc45000a41a838605aa9fd and also added a test for this new scenario. :)

packages/melos/test/commands/run_test.dart
```diff
@@ -631,6 +635,338 @@ melos run hello_script
...
+  └> analyze
+     └> RUNNING
+
+melos run analyze
+  └> dart analyze . --fatal-warnings
+     └> RUNNING
+
+Analyzing ....
+
+   info - ${currentPlatform.isWindows ? 'packages\a\main.dart' : 'packages/a/main.dart'}:3:13 - Don't invoke 'print' in production code. Try using a logging framework. - avoid_print
```
spydon, line 837: Aah the pain of Windows, this has to have two backslashes `\\`

jessicatarra: Ohhh I see, hopefully the last commit solves this issue!

jessicatarra: for some reason, even though I added the two backslashes, the error seems to be something else @spydon:

```console
              'warning - pubspec.yaml:6:11 - The path \'d:\\a\\melos\\melos\\packages\\melos\' isn\'t a POSIX-style path. Try converting the value to a POSIX-style path. - path_not_posix\n'
              '   info - packages\\a\\main.dart:3:13 - Don\'t invoke \'print\' in production code. Try using a logging framework. - avoid_print\n'
              '   info - packages\\a\\main.dart:5:10 - Missing a newline at the end of the file. Try adding a newline at the end of the file. - eol_at_end_of_file\n'
              '\n'
              '3 issues found.\n'

```

jessicatarra: @spydon I updated the test and use a `list` script instead of the `analyze` one to avoid all these issues related to windows environments, and still testing the right scenario, which is prioritazing custom scripts instead of Melos commands when scripts with same names as the commands are used. commit e1d2e9655dcf6dfc5ed7cab7217fa2a8906a8fd5

PR #694, opened 2024-04-12
fix: multiline run commands were failing

pastre:
## Description

Multiline scripts were not properly being parsed in order to run. Trimming them before creation ensures no empty spaces interfere with the script's execution

## Type of Change

- [X] 🛠️ `fix` -- Bug fix (non-breaking change which fixes an issue)
- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/scripts.dart
+++ b/packages/melos/lib/src/scripts.dart
@@ -224,7 +224,7 @@ class Script {
 
     return Script(
       name: name,
-      run: run,
+      run: run?.trim(),
       steps: steps,
       description: description,
       env: env,
--- a/packages/melos/test/commands/run_test.dart
+++ b/packages/melos/test/commands/run_test.dart
@@ -232,6 +232,63 @@ melos run hello
       );
     });
 
+    test('supports passing additional arguments to multiline scripts',
+        () async {
+      final workspaceDir = await createTemporaryWorkspace(
+        runPubGet: true,
+        configBuilder: (path) => MelosWorkspaceConfig(
+          path: path,
+          name: 'test_package',
+          packages: [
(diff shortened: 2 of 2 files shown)


spydon:
Can you give an example of a script that would fail? All our multi-line scripts work as intended as far as I know.
I think this might even break some scripts, since it would remove the newline characters and if people have escaped the newline with a backslash it would create a trailing `\` that wouldn't escape the right thing.

On another note, most of the time the new `steps` feature can be used instead of multi-line scripts:
https://melos.invertase.dev/configuration/scripts#steps

pastre:
@spydon  Sorry, I could've been a little more descriptive on the bug - and also thanks for the quick feedback!
The bug happens when passing additional arguments to a multiline YAML string - it apparently takes those arguments in as new commands and attempts to execute them
With this script:
```
scripts:
  build:
    run: |
      echo $1 $2 $3
```

I got the following output
```
$ melos build with extra arguments
melos run build
  └> echo $1 $2 $3
      with extra arguments
     └> RUNNING


ERROR: /bin/sh: line 1: with: command not found


melos run build
  └> echo $1 $2 $3
      with extra arguments
     └> FAILED
ScriptException: The script build failed to execute.
```

spydon:
@pastre ah, I see, that is filed as bug #232, and there is a good description about why these types of scripts are problematic [here](https://github.com/invertase/melos/issues/232#issuecomment-1619656025).

spydon:
I'll close this meanwhile, since this fix will break other scripts. We can continue the discussion in the issue instead.

pastre:
Cool, thanks for pointing towards the right direction @spydon :)

PR #695, opened 2024-04-12
fix!: Make `melos analyze` always use `dart analyze`

Pavel-Sulimau:
## Description

The recent `melos analyze` changes came as a breaking change surprise for our project. On our project, we do not treat info-level issues as errors because we have some `deprecated` info lints in our codebase. 

The change introduced in https://github.com/invertase/melos/pull/655 results in the call to `flutter analyze` or `dart analyze` depending on the Flutter dependency presence.

**The issue is that those commands have different default behavior and configuration options.**

 While `dart analyze` does not fail on info-level issues by default,
image

the `flutter analyze` does fail on info-level issues.
image

This results in inconsistent behavior. Also, it is not aligned with the [current documentation/API](https://melos.invertase.dev/commands/analyze#--fatal-infos) of `melos analyze`.

Since `dart analyze` is suitable for both `dart` and `flutter` packages I do not see a problem using just this command. That will fix the inconsistency and make things simpler.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-  "dart.runPubGetOnPubspecChanges": "always"
+  "dart.runPubGetOnPubspecChanges": "always",
+  "dart.lineLength": 80,
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,3 @@
 {
-  "dart.runPubGetOnPubspecChanges": "always",
-  "dart.lineLength": 80,
+  "dart.runPubGetOnPubspecChanges": "always"
 }
--- a/docs/commands/analyze.mdx
+++ b/docs/commands/analyze.mdx
@@ -20,7 +20,7 @@ melos analyze
 
 ## --fatal-infos
 Enforces a strict analysis by treating info-level issues as critical errors.
-By default, the this option is disabled.
+By default this option is disabled.
 
 ```bash
 melos analyze --fatal-infos
@@ -28,7 +28,7 @@ melos analyze --fatal-infos
 
 ## --fatal-warnings
 Enables treating warnings as fatal errors. When enabled, any warning will cause the analyzer to fail.
-By default, the this option is enabled.
+By default this option is enabled.
 
 ```bash
 melos analyze --fatal-warnings
--- a/docs/commands/analyze.mdx
+++ b/docs/commands/analyze.mdx
@@ -20,6 +20,7 @@ melos analyze
 
 ## --fatal-infos
 Enforces a strict analysis by treating info-level issues as critical errors.
+By default, the this option is disabled.
 
 ```bash
 melos analyze --fatal-infos
@@ -27,6 +28,7 @@ melos analyze --fatal-infos
 
 ## --fatal-warnings
 Enables treating warnings as fatal errors. When enabled, any warning will cause the analyzer to fail.
+By default, the this option is enabled.
 
 ```bash
 melos analyze --fatal-warnings
--- a/packages/melos/lib/src/commands/analyze.dart
+++ b/packages/melos/lib/src/commands/analyze.dart
@@ -110,10 +110,8 @@ mixin _AnalyzeMixin on _Melos {
   }) {
     final options = _getOptionsArgs(fatalInfos, fatalWarnings, concurrency);
     return <String>[
-      ..._analyzeCommandExecArgs(
-        useFlutter: workspace.isFlutterWorkspace,
-        workspace: workspace,
-      ),
+      workspace.sdkTool('dart'),
+      'analyze',
       options,
     ];
   }
@@ -161,18 +159,5 @@ mixin _AnalyzeMixin on _Melos {
       workingDirectory: package.path,
       group: group,
     );
-  }
-
-  List<String> _analyzeCommandExecArgs({
-    required bool useFlutter,
-    required MelosWorkspace workspace,
-  }) {
-    return [
-      if (useFlutter)
-        workspace.sdkTool('flutter')
-      else
-        workspace.sdkTool('dart'),
-      'analyze',
-    ];
   }
 }


docs/commands/analyze.mdx
```diff
@@ -20,13 +20,15 @@ melos analyze
 
 ## --fatal-infos
 Enforces a strict analysis by treating info-level issues as critical errors.
+By default, the this option is disabled.
```
mvolpato, line 23:
```suggestion
By default this option is disabled.
```

docs/commands/analyze.mdx
```diff
@@ -20,13 +20,15 @@ melos analyze
...
 
 ```bash
 melos analyze --fatal-infos
 ```
 
 ## --fatal-warnings
 Enables treating warnings as fatal errors. When enabled, any warning will cause the analyzer to fail.
+By default, the this option is enabled.
```
mvolpato, line 31:
```suggestion
By default this option is enabled.
```

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "dart.runPubGetOnPubspecChanges": "always"
+  "dart.runPubGetOnPubspecChanges": "always",
+  "dart.lineLength": 80,
```
spydon, line 3: This should not change in this PR, I'm not sure why this file is even checked in.

packages/melos/lib/src/commands/analyze.dart
```diff
@@ -110,10 +110,8 @@ mixin _AnalyzeMixin on _Melos {
   }) {
     final options = _getOptionsArgs(fatalInfos, fatalWarnings, concurrency);
     return <String>[
-      ..._analyzeCommandExecArgs(
-        useFlutter: workspace.isFlutterWorkspace,
-        workspace: workspace,
-      ),
+      workspace.sdkTool('dart'),
+      'analyze',
```
spydon, line 114: I'm not that familiar with tools like fvm, will they override `dart` too if they are added with `sdkPath`?
Otherwise I agree that it should be enough to just use dart here.

spydon:
Hmm, to me it sounds more consistent to make `analyze` behave like Flutter if it is a Flutter project and like Dart if it is a Dart project.

.vscode/settings.json
Pavel-Sulimau: I'm completely fine to get this one removed. The reason I added this line is that I use different line length for other dart/flutter projects of mine and it is convenient when the expected formatting settings are configured project-wide to avoid accidental missformatting by the IDE.

packages/melos/lib/src/commands/analyze.dart
Pavel-Sulimau: Yep, the `fvm ` will use the Dart version that comes with the corresponding Flutter version!

Pavel-Sulimau:
> Hmm, to me it sounds more consistent to make `analyze` behave like Flutter if it is a Flutter project and like Dart if it is a Dart project.

@spydon, I've got a couple of questions then:
1. How could I make the `melos analyze` in the current version of Melos (5.3.0) not fail on info-level issues on a Flutter (not pure Dart) package taking into account [there is no option to pass --no-fatal-infos](https://melos.invertase.dev/commands/analyze#--fatal-infos)?
2. Which advantages of using `flutter analyze` over `dart analyze` do you see nowadays? 

P.S. Some additional info on the topic:
- https://stackoverflow.com/questions/65308789/why-is-flutter-analyze-different-from-dart-analyze
- https://github.com/flutter/flutter/projects/106
- https://dart.dev/tools/dart-analyze
- https://docs.flutter.dev/reference/flutter-cli#flutter-commands

spydon:
> How could I make the melos analyze in the current version of Melos (5.3.0) not fail on info-level issues on a Flutter (not pure Dart) package taking into account [there is no option to pass --no-fatal-infos](https://melos.invertase.dev/commands/analyze#--fatal-infos)?

Ah yes, that should definitely be added as a flag to Melos!

> Which advantages of using flutter analyze over dart analyze do you see nowadays?

No advantages at all, but we shouldn't dictate from Melos side that the `dart` analyzer presets are used in a Flutter project, when people with Flutter projects will be expecting it to go through the Flutter analyzer presets. I hope that in the future the analyzer presets won't differ between Dart and Flutter.

Adding the `--no-fatal-infos` flag would also be better since it wouldn't be a breaking change.

spydon:
@Pavel-Sulimau do you want to work on adding the `--no-fatal-infos` flag instead? That should solve your problem too, right?

Pavel-Sulimau:
> @Pavel-Sulimau do you want to work on adding the `--no-fatal-infos` flag instead? That should solve your problem too, right?

@spydon, Yep, sure, I can do that! However, I would like to double-check on your suggestions as at the moment it seems to me suboptimal, i.e. adding unnecessary complexity/confusion without explicit value. Here is what is bothering me a bit about that:
1. The `dart analyze` and `flutter analyze` presets are different, so we'll have to ignore `--no-fatal-infos` for the pure Dart projects as otherwise the `dart analyze` command will fail. IMO that will be unintuitive if the `melos analyze` is going to provide such an option.  We'll also need to adjust https://melos.invertase.dev/commands/analyze to elobarate that not only `dart analyze` is used under that hood.
2. [Flutter repository itself already uses](https://github.com/flutter/samples/blob/f4083c19b21d723cc9973eed24d289e1022b5aa2/tool/flutter_ci_script_shared.sh#L15C9-L15C21) `dart analyze` for Flutter projects, some other well-known packages like flutter_bloc ([see this](https://github.com/felangel/bloc/blob/45591078de7e0d5d3b03d4488e42c4be020ae738/.github/actions/flutter_package/action.yaml#L64C12-L64C24)) and plus_plugins ([see this](https://github.com/fluttercommunity/plus_plugins/blob/15985a1d1283122826aeff071473aed06ccf0265/.github/workflows/all_plugins.yaml#L37C39-L37C52)) do that as well.

So, could you please confirm that using both `flutter analyze` and `dart analyze` and finding the common denominator between their presets for `melos analyze` is the preferred way to go?

spydon:
>     1. The `dart analyze` and `flutter analyze` presets are different, so we'll have to ignore `--no-fatal-infos` for the pure Dart projects as otherwise the `dart analyze` command will fail. IMO that will be unintuitive if the `melos analyze` is going to provide such an option.  We'll also need to adjust https://melos.invertase.dev/commands/analyze to elobarate that not only `dart analyze` is used under that hood.

We just don't add `--fatal-infos` for dart projects when `--no-fatal-infos` is present, so it will behave like one would expect.

>     2. [Flutter repository itself already uses](https://github.com/flutter/samples/blob/f4083c19b21d723cc9973eed24d289e1022b5aa2/tool/flutter_ci_script_shared.sh#L15C9-L15C21) `dart analyze` for Flutter projects, some other well-known packages like flutter_bloc ([see this](https://github.com/felangel/bloc/blob/45591078de7e0d5d3b03d4488e42c4be020ae738/.github/actions/flutter_package/action.yaml#L64C12-L64C24)) and plus_plugins ([see this](https://github.com/fluttercommunity/plus_plugins/blob/15985a1d1283122826aeff071473aed06ccf0265/.github/workflows/all_plugins.yaml#L37C39-L37C52)) do that as well.
> 
> 
> So, could you please confirm that using both `flutter analyze` and `dart analyze` and finding the common denominator between their presets for `melos analyze` is the preferred way to go?

Yeah, there are definitely flutter projects using the Dart analyzer too.

I have a feeling that the big majority is using `flutter analyze` for their Flutter apps though, but if that is not the case, I think we can go your way and just just `dart analyze`. I made a little poll [here](https://twitter.com/spydon/status/1779613358339084616) to check, so hopefully we get a lot of data on that one.

spydon:
We had a chat internally, and people are for just using `dart analyze`.
And also, the poll came out on like 40/60 (with 100 participants), so there wasn't really a clear "winner" there anyways.
So let's go ahead with this PR!
I guess we'll have to make it a breaking change though, since it is will change the current behavior for Flutter projects.

Pavel-Sulimau:
> We had a chat internally, and people are for just using `dart analyze`. And also, the poll came out on like 40/60 (with 100 participants), so there wasn't really a clear "winner" there anyways. So let's go ahead with this PR! I guess we'll have to make it a breaking change though, since it is will change the current behavior for Flutter projects.

All right, thanks for reaching out to people Lukas! 
Let me know then if there is anything else I need to improve/fix in the current PR.

spydon:
> > We had a chat internally, and people are for just using `dart analyze`. And also, the poll came out on like 40/60 (with 100 participants), so there wasn't really a clear "winner" there anyways. So let's go ahead with this PR! I guess we'll have to make it a breaking change though, since it is will change the current behavior for Flutter projects.
> 
> All right, thanks for reaching out to people Lukas! Let me know then if there is anything else I need to improve/fix in the current PR.

I think it looks good as it is, I'll go ahead and merge it now. :)
I think we'll wait a bit with making a release to see if there is anything more coming in with potential breaking changes.

Issue #675, opened 2024-03-21
fix:  melos overrides the applied filter for dependents/dependencies


xsahil03x:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Version

5.2.1

### Description

Currently, the order in which melos exec calculates the packages are as follows:

https://github.com/invertase/melos/blob/main/packages/melos/lib/src/package.dart#L597-L628
1.  `filters` (eg: applyIgnore, applyDirExists, applyFileExists)
2. `applyDiff`
3. `applyIncludeDependentsOrDependencies` if enabled

Because we are executing `applyIncludeDependentsOrDependencies` at the last step, the added packages skips all the filters applied on the first step. So even tho a package is skipped in the first step, it can re-appear through the last `applyIncludeDependentsOrDependencies` step.

### Steps to reproduce

Suppose we have a monorepo with 2 packages.

<table>
<tr>
 <td> Package
 <td> Dependencies
 <td> Have Changes
 <td> Contains Test
<tr>
 <td> Package A
 <td> Package B
 <td> false
 <td> false
<tr>
 <td> Package B
 <td> -
 <td> true
 <td> true
</table>

Now if we try to run the command:

```melos exec --fail-fast --dir-exists="test" --diff=origin/main...HEAD -c 1 --include-dependents -- flutter test```

Ideally the command should only run for `Package B` as `Package A` contains no test. But in the current implementation it will also include `Package A` as it depends on `Package B`.

### Expected behavior

The `filters` should also be applied on `dependents/dependencies`.

**Comment by spydon** [comment]
```
    --include-dependents                         Include all transitive dependents for each package that matches the other filters. The included packages skip --ignore and --diff checks.
```

This is how it is intended to work, and also how the docs describe it.

xsahil03x:
> ```
> --include-dependents    Include all transitive dependents for each package that matches the other filters. The included packages skip --ignore and --diff checks.
> ```
> 
> This is not a bug, it is how it is intended to work, and also how the docs describe it.

Thanks, i missed.

Pavel-Sulimau:
Hey @xsahil03x, I ended up implementing the approach of setting the global `MELOS_PACKAGES` filter with the values from `melos list --diff=origin/main...HEAD --include-dependents` in the CI pipeline for our big project. 

The approach is described in https://itnext.io/flutter-selective-ci-checks-2d79ffbd26e5 article. Hopefully, it may be a workaround for you as well!

PR #702, opened 2024-04-18
feat: implement reversed-filter-group-priority option

Pavel-Sulimau:
## Description

This PR adds a new filtering option that implements the problem/desired behavior described in [this issue](https://github.com/invertase/melos/issues/675).

The implementation is backward compatible, so the default filtering behavior should stay as it is.

Ideally, I'd love to have the possibility to prioritize filters based on the order of their input, but it seems to me that would require way more changes, so this is the simpler trade-off solution I have come up with so far.

---

The alternative approach I ended up implementing in the CI workflow of our big project implies setting the global `MELOS_PACKAGES` scope filter with the values from `melos list --diff=origin/main...HEAD --include-dependents`. The approach is described in https://itnext.io/flutter-selective-ci-checks-2d79ffbd26e5 article. Hopefully, it is helpful for other looking for optimization in their CI pipelines!

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/docs/filters.mdx
+++ b/docs/filters.mdx
@@ -5,9 +5,21 @@ description: Learn more about all the package filtering flags in Melos.
 
 # Filtering Packages
 
-Each Melos command can be used alongside the following global filters:
+Each Melos command can be used alongside the following global filters.
 
-## --no-private
+The filters are divided into 2 priority groups. The priority determines the
+order of their application. The grouping has been introduced with backward
+compatibility in mind. It is possible to reverse the order of the priority
+groups to achive different filtering result.
+
+## --reversed-filter-group-priority
+Reverses the priority of filtering groups. Useful for applying diff-based
+filtering first, allowing the filters from the other group to further narrow
+down the packages.
+
+## 1st Priority Group
+
+### --no-private
 
 Exclude private packages (`publish_to: none`). They are included by default.
 
@@ -15,7 +27,7 @@ Exclude private packages (`publish_to: none`). They are included by default.
 melos bootstrap --no-private
 ```
 
-## --published
+### --published
 
 Filter packages where the current local package version exists on pub.dev.
 
@@ -26,7 +38,7 @@ melos bootstrap --published
 Use `--no-published` to filter packages that have not had their current version
 published yet.
 
-## --scope
+### --scope
 
 Include only packages with names matching the given glob. This option can be
 repeated.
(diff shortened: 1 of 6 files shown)


Pavel-Sulimau:
Hey @Salakar, @spydon, your feedback on the proposed changes would be greatly appreciated. Thank you in advance!

spydon:
> Hey @Salakar, @spydon, your feedback on the proposed changes would be greatly appreciated. Thank you in advance!

I'm still in the US, but I should have time to look at this on the weekend again! I'm still a bit unsure of the reverse concept and priority groups.

spydon:
I had a look again and I agree with what it is trying to do, but as I've said before I don't like the complexity of introducing the priority groups. I haven't come up with any other ideas though, do you have any other ideas how this could be handled?

Pavel-Sulimau:
@spydon, the most obvious and prominent idea in my opinion is to implement filters' prioritization based on the order of their input, however, I foresee this feature will require quite some changes.

Alternatively, we could just create a flag (how would we call it?) specifically handling this case to prioritize `applyDiff` and `applyIncludeDependentsOrDependencies`, but it looks like a less generic solution.

So, I agree that the priority groups is not the greatest way to solve the case, but I don't see many other ways at the moment how to do that elegantly without much effort/potential to break what is already working.

spydon:
> @spydon, the most obvious and prominent idea in my opinion is to implement filters' prioritization based on the order of their input, however, I foresee this feature will require quite some changes.

That sounds like a great idea! If some filters aren't coming in through flags, but through config, we have to have a way to decide the order between the flags and those ones too right?

Pavel-Sulimau:
> If some filters aren't coming in through flags, but through config, we have to have a way to decide the order between the flags and those ones too right?

I believe it should be possible.

spydon:
> > If some filters aren't coming in through flags, but through config, we have to have a way to decide the order between the flags and those ones too right?
> 
> I believe it should be possible.

Do you want to dig deeper into that option? I think it would be a better solution to the problem.

Pavel-Sulimau:
I would love to, however as I noted in the PR description it will likely require way more work for which I do not have enough time in the upcoming month or 2, so I'll leave it like this in the case smth has time and desire to implement the other approach, otherwise I can get back to it myself sometime later.

spydon:
Since we never managed to figure out a good way to do this I'll close this for now.

PR #708, opened 2024-04-25
fix: tryParse line-length to int when it's not already an integer

jkoenig134:
## Description

The flag `--line-length` was added, but it seems like it doesn't work as ArgParser doesn't seem to transform this param to int by default. This fix takes the `String` argument and tries parsing it to an `int`.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/command_runner/format.dart
+++ b/packages/melos/lib/src/command_runner/format.dart
@@ -38,6 +38,7 @@ class FormatCommand extends MelosCommand {
     final concurrency = int.parse(argResults!['concurrency'] as String);
     final lineLength = switch (argResults?['line-length']) {
       final int length => length,
+      final String length => int.tryParse(length),
       _ => null,
     };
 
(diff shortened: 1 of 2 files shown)


spydon (approved):
Lgtm, it would be good with a test for this though

utamori:
@jkoenig134 thank you!
I tried to add a test, but it didn't work.

I just made the following modification and was able to confirm that it works correctly at hand.

```dart
    final lineLength =
        switch (int.tryParse(argResults?['line-length'] as String)) {
      final int length => length,
      _ => null,
    };
```

melos activate
melos foramt --line-length 120 --set-exit-if-changed

jkoenig134:
@spydon added the test

spydon (approved):
Super, thanks!

jkoenig134:
@spydon as a follow-up i would really like to add setExitIfChanged and lineLength to the config.

I am mostly though with that. As it is not only a Bugfix - can I PR this?

spydon:
> @spydon as a follow-up i would really like to add setExitIfChanged and lineLength to the config.
> 
> I am mostly though with that. As it is not only a Bugfix - can I PR this? 

Sure, sounds good!

Issue #470, opened 2023-02-08
fix:  [Windows] Multiple command with `melos run` does not work


ferraridamiano:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Version

2.9.0

### Description

On Windows if you run `melos run my_script` and `my_script` contains multiple commands it will only be executed the first command. It works properly on Linux

### Steps to reproduce

Example script:

```yaml
scripts:
  my_script:
    run: |
      command1
      command2
    description: Script description
```

If you want to try on a real repo run `melos run compile_icons` on [this repo](https://github.com/ferraridamiano/ConverterNOW) (you need to `dart pub global activate vector_graphics_compiler` before)

### Expected behavior

All the commands of the script are executed on Windows

matthew-james-wooliesx:
This is making it difficult for us to use `melos` on a project with Windows- and MacOS-based devs.

tomwyr:
Same situation in the project I work on. There are scripts consisting of 4-5 sub-scripts I need to run one by one in correct order.

JesperBllnbm:
As a workaround on Windows you can use a single ampersand (&) to separate multiple commands on one command line. When a single ampersand is used, cmd.exe runs the first command, and then the second command.

`command1 & command2`

spydon:
The new `steps` feature solved this: https://melos.invertase.dev/configuration/scripts#steps

JesperBllnbm:
Cool! 

But on Melos 5.2.2. I get:

`melos.yaml: The property run at scripts/pre-commit is required but missing`

when using the example from the documentation. Is the feature not yet available?

timcreatedit:
> The new `steps` feature solved this: https://melos.invertase.dev/configuration/scripts#steps

I like the `steps` feature, but to my understanding it doesn't really fix the issue. One major drawback is that steps seem to all run in the workspace root, so anything where you want to change the working dir in between requires the old `&&` workaround again. See #705

PR #711, opened 2024-05-13
fix: maintain working directory across script steps

jessicatarra:
This PR addresses the issue raised in #470, where the working directory is not maintained across multiple steps in a script.

For example, consider the following script:

```yaml
fail:
  steps:
    - cd packages/
    - pwd
    
```
When running this script using melos run, the output will show the workspace root directory, not the `packages/` directory as expected.

This PR introduces a change to ensure that the working directory is properly maintained across multiple steps in a script.

The key changes are:

- Modify the script execution logic to preserve the working directory from one step to the next.
- Add a safeguard to prevent users from mixing && operators with multi-step scripts, as this can lead to unexpected behavior.

By running the script defined in the issue with the fix implemented in this PR, the output is now:

```shell
➜  melos git:(bugfix/melos_run_script_steps_working_directory) ✗ melos success
Building package executable...
Built melos:melos.
melos run success
  └> cd packages/
     └> RUNNING


melos run success
  └> cd packages/
     └> SUCCESS

melos run success
  └> pwd
     └> RUNNING

/Users/jessicatarra/Development/melos/packages

melos run success
  └> pwd
     └> SUCCESS
```

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -278,6 +278,7 @@ mixin _RunMixin on _Melos {
     final shell = PersistentShell(
       logger: logger,
       workingDirectory: config.path,
+      environment: environment,
     );
 
     await shell.startShell();
--- a/packages/melos/lib/src/commands/run.dart
+++ b/packages/melos/lib/src/commands/run.dart
@@ -275,48 +275,24 @@ mixin _RunMixin on _Melos {
     Script script,
     Map<String, String> environment,
   ) async {
-    for (final step in steps) {
-      final scriptCommand = _buildScriptCommand(step, scripts);
-
-      final scriptSourceCode = targetStyle(
-        step.withoutTrailing('\n'),
-      );
-
-      await _executeAndLogCommand(
-        script,
-        scriptSourceCode,
-        scriptCommand,
-        environment,
-      );
-    }
-  }
-
-  Future<void> _executeAndLogCommand(
-    Script script,
-    String scriptSourceCode,
-    String scriptCommand,
-    Map<String, String> environment,
-  ) async {
-    logger.command('melos run ${script.name}');
-    logger.child(scriptSourceCode).child(runningLabel).newLine();
-
-    final exitCode = await startCommand(
-      [scriptCommand],
+    final shell = PersistentShell(
       logger: logger,
-      environment: environment,
       workingDirectory: config.path,
     );
 
-    logger.newLine();
+    await shell.startShell();
     logger.command('melos run ${script.name}');
-    final resultLogger = logger.child(scriptSourceCode);
 
-    if (exitCode != 0) {
-      resultLogger.child(failedLabel);
-    } else {
-      resultLogger.child(successLabel);
+    for (final step in steps) {
+      final scriptCommand = _buildScriptCommand(step, scripts);
+
+      final shouldContinue = await shell.sendCommand(scriptCommand);
+      if (!shouldContinue) {
+        break;
+      }
     }
-    logger.newLine();
+
+    await shell.stopShell();
   }
 }
 
--- a/packages/melos/lib/src/common/persistent_shell.dart
+++ b/packages/melos/lib/src/common/persistent_shell.dart
@@ -3,17 +3,20 @@ import 'dart:convert';
 import 'dart:io';
 
 import '../logging.dart';
+import 'environment_variable_key.dart';
 import 'platform.dart';
 import 'utils.dart';
 
 class PersistentShell {
   PersistentShell({
     required this.logger,
+    required this.environment,
     this.workingDirectory,
   });
 
   final _isWindows = currentPlatform.isWindows;
   final MelosLogger logger;
+  final Map<String, String> environment;
   final String? workingDirectory;
   late final Process _process;
   Completer<void>? _commandCompleter;
@@ -27,6 +30,10 @@ class PersistentShell {
       executable,
       [],
       workingDirectory: workingDirectory,
+      environment: {
+        ...environment,
+        EnvironmentVariableKey.melosTerminalWidth: terminalWidth.toString(),
+      },
     );
 
     _listenToProcessStream(_process.stdout);
--- a/packages/melos/lib/src/common/persistent_shell.dart
+++ b/packages/melos/lib/src/common/persistent_shell.dart
@@ -1,0 +1,101 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import '../logging.dart';
+import 'platform.dart';
+import 'utils.dart';
+
+class PersistentShell {
+  PersistentShell({
+    required this.logger,
+    this.workingDirectory,
+  });
+
+  final _isWindows = currentPlatform.isWindows;
+  final MelosLogger logger;
+  final String? workingDirectory;
+  late final Process _process;
+  Completer<void>? _commandCompleter;
+  final String _successEndMarker = '__SUCCESS_COMMAND_END__';
+  final String _failureEndMarker = '__FAILURE_COMMAND_END__';
+
+  Future<void> startShell() async {
+    final executable = _isWindows ? 'cmd.exe' : '/bin/sh';
+
+    _process = await Process.start(
+      executable,
+      [],
+      workingDirectory: workingDirectory,
+    );
+
+    _listenToProcessStream(_process.stdout);
+    _listenToProcessStream(_process.stderr, isError: true);
+  }
+
+  Future<bool> sendCommand(String command) async {
+    assert(_commandCompleter == null, 'A command is already in progress.');
+    _commandCompleter = Completer<void>();
+
+    final fullCommand = _buildFullCommand(command);
+    _process.stdin.writeln(fullCommand);
+
+    return _awaitCommandCompletion();
+  }
+
+  Future<void> stopShell() async {
+    await _process.stdin.close();
+    final exitCode = await _process.exitCode;
+    if (exitCode == 0) {
+      logger.log(successLabel);
+      return;
+    }
+    logger.log(failedLabel);
+  }
+
+  Future<bool> _awaitCommandCompletion() async {
+    try {
+      await _commandCompleter!.future;
+      return true;
+    } catch (e) {
+      return false;
+    } finally {
+      _commandCompleter = null;
+    }
+  }
+
+  void _listenToProcessStream(
+    Stream<List<int>> stream, {
+    bool isError = false,
+  }) {
+    stream.listen((event) {
+      final output = utf8.decode(event, allowMalformed: true);
+      logger.logAndCompleteBasedOnMarkers(
+        output,
+        _successEndMarker,
+        _failureEndMarker,
+        _commandCompleter,
+        isError: isError,
+      );
+    });
+  }
+
+  String _buildFullCommand(String command) {
+    final formattedScriptStep =
+        targetStyle(command.addStepPrefixEmoji().withoutTrailing('\n'));
+    final echoCommand = 'echo "$formattedScriptStep"';
+    final echoSuccess = 'echo $_successEndMarker';
+    final echoFailure = 'echo $_failureEndMarker';
+
+    if (_isWindows) {
+      return '''
+      $echoCommand && $command || VER>NUL && if %ERRORLEVEL% NEQ 0 ($echoFailure) else ($echoSuccess)
+    ''';
+    }
+
+    return '''
+   $echoCommand && $command || true && if [ \$? -ne 0 ]; 
+    then $echoFailure; else $echoSuccess; fi
+  ''';
+  }
+}
--- a/packages/melos/lib/src/common/utils.dart
+++ b/packages/melos/lib/src/common/utils.dart
@@ -73,6 +73,10 @@ final melosPackageUri = Uri.parse('package:melos/melos.dart');
 final _camelCasedDelimiterRegExp = RegExp(r'[_\s-]+');
 
 extension StringUtils on String {
+  String addStepPrefixEmoji() {
+    return '➡️ step: $this';
+  }
+
   String indent(String indent) {
     final split = this.split('\n');
 
--- a/packages/melos/lib/src/commands/runner.dart
+++ b/packages/melos/lib/src/commands/runner.dart
@@ -29,6 +29,7 @@ import '../common/glob.dart';
 import '../common/intellij_project.dart';
 import '../common/io.dart';
 import '../common/pending_package_update.dart';
+import '../common/persistent_shell.dart';
 import '../common/platform.dart';
 import '../common/utils.dart' as utils;
 import '../common/utils.dart';
--- a/packages/melos/lib/src/logging.dart
+++ b/packages/melos/lib/src/logging.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:ansi_styles/ansi_styles.dart';
 import 'package:cli_util/cli_logging.dart';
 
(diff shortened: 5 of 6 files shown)


packages/melos/lib/src/commands/run.dart
```diff
@@ -287,7 +295,12 @@ mixin _RunMixin on _Melos {
         scriptSourceCode,
         scriptCommand,
         environment,
+        currentWorkingDirectory,
       );
+
+      if (step.contains('cd ')) {
```
spydon, line 301: I think there must be a better way to keep the environment between steps than to look for cd commands 🤔
`cd` is also not the only way to change directory in Linux, for example `pushd` and `popd` are quite commonly used in scripts.

And I guess other things in the environment should be taken into consideration too, like exporting variables for example.

jessicatarra: It makes sense, I'll think of another approach then. Thanks for the feedback!

Salakar: what about if a step and/or script could define it's working directory rather than us trying to detect it and magically preserve it perhaps, e.g.:

```yaml
fail:
  # For all steps:
  # workingDirectory: packages/
  steps:
    - run: pwd
      # for an individual step and also overrides top level if set
      # workingDirectory: packages/
    - ls -la # still support just a string item in the steps array
 ```

Comment (jessicatarra) [reply packages/melos/lib/src/commands/run.dart:301]: Nice! I'm all in, it makes a tons of sense. I will start working on it

Comment (spydon) [reply packages/melos/lib/src/commands/run.dart:301]: @salakar I think that many of the people that have asked for this feature don't just need to be in one directory, but jump around during the multi-step script. Isn't it possible to keep the shell session alive and just keep on executing all steps in there, instead of creating a new shell for each step? 🤔

Comment (jessicatarra) [reply packages/melos/lib/src/commands/run.dart:301]: I did a proof of concept of the approach mentioned by @spydon, and it's doable. It actually follows the same logic of GitHub workflows, also mentioned in the comment replied by @timcreatedit:

> I would understand if you keep it like this (also because it's more consistent with GitHub workflows)

The code that I used for the PoC is as follows:

```dart

class PersistentShell {
  PersistentShell({required bool isWindows}) {
    _isWindows = isWindows;
  }
  late Process _process;
  late bool _isWindows;

  Future<void> startShell() async {
    _process = await Process.start(
      _isWindows ? 'cmd.exe' : '/bin/sh',
      _isWindows ? ['/Q'] : [],
    );

    _process.stdout.transform(utf8.decoder).listen(print);
    _process.stderr.transform(utf8.decoder).listen((data) {
      print('Error: $data');
    });
  }

  void sendCommand(String command) {
    _process.stdin.writeln(_isWindows ? command : 'eval "$command"');
  }

  Future<void> stopShell() async {
    await _process.stdin.close();
    await _process.exitCode;
  }
}

void main() async {
  var shell = PersistentShell(isWindows: Platform.isWindows);
  await shell.startShell();
  shell.sendCommand('echo Hello World');
  await shell.stopShell();
}

```
I implemented the code and it works really well, as the hardcoded condition to check if 'cd ' is part of the step command but way better, because it's dynamic. Let me know if this is what you meant @spydon

spydon: Sorry for the late reply, this is what I meant indeed! :)

spydon:
Any progress on this @jessicatarra? :blush:

jessicatarra:
Hey everyone,
@spydon  Sorry for the late reply. Work's been pretty busy lately, but I finally got some time this weekend to dig back into this fix. I managed to make things work, but now I'm stuck on figuring out how to keep the logs consistent with our current style.
The issue is, this persistent shell (as it is implemented) makes it tricky to track the results of individual commands. To help us brainstorm a solution, I'm sharing my implementation so far. Let's see what we can come up with together.

## *Example script*

```
melos:test:
  steps:
    - cd packages/melos
    - pwd
    - dart test .
```
  
## **Output**
```
melos run melos:test
➡️ step: cd packages/melos
➡️ step: pwd
/Users/jessicatarra/Development/melos/packages/melos
➡️ step: dart test .

… (long logs from dart test step)  
ERROR: Shell process completed with errors.
Shell process completed some steps successfully.  
```

packages/melos/lib/src/common/utils.dart
```diff
@@ -205,6 +209,77 @@ class PromptException extends MelosException {
...
+    required MelosLogger logger,
+    String? workingDirectory,
+  }) {
+    _logger = logger;
+    _workingDirectory = workingDirectory;
+  }
+
+  final _isWindows = currentPlatform.isWindows;
+  late final MelosLogger _logger;
+  late final String? _workingDirectory;
```
spydon, line 223: These don't have to be private, they can be set directly in the constructor like this:
```dart
  PersistentShell({
    required MelosLogger logger,
    String? workingDirectory,
  }) : _logger = logger,  _workingDirectory = workingDirectory;
  ```
  
  But I don't think there is any reason to keep them private, so you can simplify it even further:
  ```dart
  PersistentShell({
    required this.logger,
    this.workingDirectory,
  });
  ```

File: packages/melos/lib/src/common/utils.dart [hunk 3]
```diff
@@ -205,6 +209,77 @@ class PromptException extends MelosException {
   String toString() => 'Was unable to prompt for input:\n$message';
 }
 
+class PersistentShell {
```
spydon, line 212: This should live in its own file instead of in the utils file :)
I think we can place it in `common` too.

packages/melos/lib/src/common/utils.dart
```diff
@@ -205,6 +209,77 @@ class PromptException extends MelosException {
...
+    String? workingDirectory,
+  }) {
+    _logger = logger;
+    _workingDirectory = workingDirectory;
+  }
+
+  final _isWindows = currentPlatform.isWindows;
+  late final MelosLogger _logger;
+  late final String? _workingDirectory;
+  late Process _process;
```
spydon, line 224: I guess this can be final?

packages/melos/lib/src/common/utils.dart
```diff
@@ -205,6 +209,77 @@ class PromptException extends MelosException {
...
+      command.addStepPrefixEmoji().withoutTrailing('\n'),
+    );
+
+    final echoCommand = 'echo "$formattedScriptStep"';
+
+    final fullCommand = _isWindows
+        ? '$echoCommand && $command'
+        : 'eval "$echoCommand && $command"';
+
+    _process.stdin.writeln(fullCommand);
```
spydon, line 274: I think after each command has run we have to check the exit code, I'm not sure how we would do that though.
The only easy way I can think of is to store the exit code to a file, but that feels like a very inefficient way...

spydon: I guess since all of `stdout` is captured we could also echo the error codes and just not print them to the user,
so `echo $?` for *nix and `echo %ERRORLEVEL%` for windows and then just stop the process if a non-zero result is detected.

packages/melos/lib/src/common/utils.dart
```diff
@@ -205,6 +209,77 @@ class PromptException extends MelosException {
...
+      },
+      onDone: () {
+        /// TODO: Identify and log the specific steps that have completed
+        /// with errors.
+        _logger.error('Shell process completed with errors.');
+      },
+    );
+  }
+
+  void sendCommand(String command) {
```
spydon, line 261: We probably need to make this `Future<void>` and add a completer that we complete from the listeners (when we get the result code, more about that further down), and also we can add an assertion so that `sendCommand` isn't called when there already is one in progress.

spydon (commented):
> @spydon Sorry for the late reply. Work's been pretty busy lately, but I finally got some time this weekend to dig back into this fix.

No worries! Thanks for looking into it again. :)

> I managed to make things work, but now I'm stuck on figuring out how to keep the logs consistent with our current style.

I left a few suggestions that I think might help, let me know what you think.

packages/melos/lib/src/common/utils.dart
jessicatarra: It makes sense, and it also worked out. Thank you for the suggestion. As for stopping the process if a non-zero result is detected, I'm not entirely sure about it, as that's not the way it's working right now. However, if the you guys think is a good idea, we can include it, perhaps as a flag (?). If the user intentionally passes this flag, then the process will stop

packages/melos/lib/src/common/utils.dart
jessicatarra: I tried this approach, however, it creates a lot of noise in the previous tests related to the 'multiple steps script' functionality. And it's seems safe to call `sendCommand` without the assert.
Additionally, speaking about tests, I needed to update the asset part of most of them (related to multiple steps script). This is because the way logs work in the persistent shell has changed significantly, and I'm not entirely sure how to approach this. I wasn't able to figure out how to combine the current logs with the way the persistent shell is implemented. Any suggestion? 

Any advice on how to handle this issue would be greatly appreciated.

packages/melos/lib/src/common/utils.dart
spydon: I think the user should be responsible for letting a command fail and then continue, with a step ending in `|| true` for example. This is how most things like this work as far as I'm aware. :)

packages/melos/lib/src/common/utils.dart
spydon: What noise is that approach producing? Result codes should be intercepted in the listener so that they are not printed back to the user.

For the return type of `sendCommand` isn't it fire and forget if it doesn't use the completer? And hence you can get race conditions, like queueing up a command before the previous one has given back a result code?

jessicatarra: I tried a new approach, and it worked out and did not impact the current tests. I think my first attempt was quite different from the one you suggested. Take a look at my last commit and let me know :)

packages/melos/lib/src/commands/run.dart
```diff
@@ -275,48 +275,21 @@ mixin _RunMixin on _Melos {
...
-    final resultLogger = logger.child(scriptSourceCode);
 
-    if (exitCode != 0) {
-      resultLogger.child(failedLabel);
-    } else {
-      resultLogger.child(successLabel);
+    for (final step in steps) {
+      final scriptCommand = _buildScriptCommand(step, scripts);
+
+      await shell.sendCommand(scriptCommand);
```
spydon, line 289: So on error here it should break out of the loop

packages/melos/lib/src/common/persistent_shell.dart
```diff
@@ -0,0 +1,85 @@
...
+      executable,
+      _isWindows ? ['/C'] : [],
+      workingDirectory: workingDirectory,
+    );
+
+    _process.stdout.listen(
+      (event) {
+        final output = utf8.decode(event, allowMalformed: true);
+        _handleCommandCompletion(output);
+        logger.logWithoutNewLine(output);
```
spydon, line 34: I think we can omit printing the labels here and just print the actual command output

packages/melos/lib/src/common/utils.dart
spydon: Yes that's pretty much how I imagined it!

packages/melos/lib/src/common/persistent_shell.dart
jessicatarra: by labels you mean 'success', 'running' or 'failed'?

spydon: Exactly, and then at then end of all steps we can print whether it was a success or failed instead

packages/melos/lib/src/common/persistent_shell.dart
```diff
@@ -0,0 +1,84 @@
...
+    _process = await Process.start(
+      executable,
+      _isWindows ? ['/C'] : [],
+      workingDirectory: workingDirectory,
+    );
+
+    _process.stdout.listen((event) {
+      final output = utf8.decode(event, allowMalformed: true);
+      logger.logWithoutNewLine(output);
+      _commandCompleter?.complete();
```
spydon, line 33: It's not necessarily complete here right? This will currently complete after the first line has been send to stdout I think?

jessicatarra: I did some tests, and without this, some of them failed due to `TimeoutException after 0:00:30.000000: Test timed out after 30 seconds.`.
 
The flow is as follows: when a command is sent using `sendCommand`, the `_commandCompleter` is initialized then the stdout listener completes the `_commandCompleter` when it receives output, indicating that the command has finished executing. A similar process occurs for the stderr listener. 

The wait occurs in the `_awaitCommandCompletion` method.

spydon: But what I mean is that you might hit this callback multiple times for a command right, if the command prints multiple lines? And in those cases you only want the finish the completer for the last line.

jessicatarra: Hmm, I see your point now. One way to achieve this is to have some kind of marker to be notified if the command hits the last line. For example:

```dart
  final String _successEndMarker = '__SUCCESS_COMMAND_END__';
  final String _failureEndMarker = '__FAILURE_COMMAND_END__';
  
    void _handleCommandCompletion(String output) {
    if (output.contains(_successEndMarker)) {
      _commandCompleter?.complete();
    } else if (output.contains(_failureEndMarker)) {
      _commandCompleter?.completeError('Command failed');
    }
  }
```

And then, 

```dart
  String _buildFullCommand(String command) {
    final formattedScriptStep =
        targetStyle(command.addStepPrefixEmoji().withoutTrailing('\n'));

    final echoCommand = 'echo "$formattedScriptStep"';
    final echoSuccess = 'echo $_successEndMarker';
    final echoFailure = 'echo $_failureEndMarker';

    if (_isWindows) {
      return '''
        $echoCommand && $command && if %ERRORLEVEL%==0 
            ($echoSuccess) else ($echoFailure || true)
      ''';
    }

    return '''
      eval "$echoCommand && $command && if [ \$? -eq 0 ]; 
      then $echoSuccess; else $echoFailure; fi || true"
    ''';
  }

```


The only problem is that this will be printed out. To circumvent this, we can print the report of all commands only when the `stopShell` method is triggered. Additionally, if the user wants to see all the logs immediately, we could add a `verbose` flag.

spydon: Isn't the output handled by `logWithoutNewLine`? So you could just skip that one if it detects a line with the markers?

spydon: @jessicatarra any updates on this? :)

jessicatarra: Yes! I was able to handle this by using a new method very similar to the `logWithoutNewLine` function. This new method will detect the marker, whether it's a success or failure, and also complete the completer.

packages/melos/lib/src/commands/run.dart
```diff
@@ -275,48 +275,24 @@ mixin _RunMixin on _Melos {
...
-    Map<String, String> environment,
-  ) async {
-    logger.command('melos run ${script.name}');
-    logger.child(scriptSourceCode).child(runningLabel).newLine();
-
-    final exitCode = await startCommand(
-      [scriptCommand],
+    final shell = PersistentShell(
       logger: logger,
-      environment: environment,
```
spydon, line 306: We need to set the environment in the new shell right? So that it isn't missing set environment variables etc, right?

spydon (commented):
Starts to look very good!
No idea about the test failing on windows though...

packages/melos/lib/src/commands/run.dart
jessicatarra: Right, I just added it as a parameter to initialize the persistent shell

spydon:
I see that there are some tests failing for windows on `main` too, is it maybe the same ones @jessicatarra, and we just have bad tests that aren't affected by this PR?

jessicatarra:
That's a good point @spydon . I'll check that out. Maybe it would be a good idea to separate the work and open a new PR to fix the failing Windows tests.

spydon:
> That's a good point @spydon . I'll check that out. Maybe it would be a good idea to separate the work and open a new PR to fix the failing Windows tests.

Yeah, if there aren't tests failing due to this PR that sounds like a good idea

spydon (approved):
Lgtm! @jessicatarra this is ready for merge now right?

spydon (approved):

PR #751, opened 2024-08-19
test: fix tests in Windows environment

jessicatarra:
## Description

Hello Everyone 👋. Some tests were failing when running in Windows environments, so this PR addresses this issue and offers a fix.

Fixes: #757 

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -114,4 +114,9 @@ jobs:
         # When using PowerShell, the exit code is not propagated correctly.
         # Even running `CMD /C melos ...` from PowerShell does not work.
         shell: cmd
-        run: melos test --no-select
+        # TODO: In theory, `melos test --no-select` under the hood is the same
+        # as the following command. For some reason, using this one directly
+        # provides the actual exit code of the process, avoiding
+        # always returning 'SUCCESS' even when tests are failing.
+        run: melos exec --dir-exists=test --concurrency 1 -- "dart test"
+
--- a/packages/melos/test/commands/format_test.dart
+++ b/packages/melos/test/commands/format_test.dart
@@ -111,7 +111,24 @@ $ melos format
         stderrEncoding: utf8,
       );
 
-      expect(result.exitCode, equals(1));
+      ///TODO: exit code in windows environments is not reliable when is not
+      ///equal to 0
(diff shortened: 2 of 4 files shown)


jessicatarra:
@spydon even though this fix addresses the issue, it's a very hacky solution and I'm still testing in order to figure out the root cause of the problem

.github/workflows/validate.yaml
```diff
@@ -114,4 +114,9 @@ jobs:
         # When using PowerShell, the exit code is not propagated correctly.
         # Even running `CMD /C melos ...` from PowerShell does not work.
         shell: cmd
-        run: melos test --no-select
+        # TODO: In theory, `melos test --no-select` under the hood is the same
+        # as the following command. For some reason, using this one directly
+        # provides the actual exit code of the process, avoiding
+        # always returning 'SUCCESS' even when tests are failing.
```
spydon, line 120: Sounds like something we should file a bug for

packages/melos/test/commands/format_test.dart
```diff
@@ -46,6 +48,7 @@ void main() {
 
       logger = TestLogger();
       final config = await MelosWorkspaceConfig.fromWorkspaceRoot(workspaceDir);
+      expectedExitCode = currentPlatform.isWindows ? 0 : 1;
```
spydon, line 51: Why is the expected exit code 0 for windows? :thinking:

jessicatarra: Exactly, that's what I'm trying to figure out. I'm not really sure what's going on. It's the same issue in the `run_test` and even the `windows_test` step in the pipeline have this problem, exit code it's 1 but is 0 in windows environments.

.github/workflows/validate.yaml
jessicatarra: Yes, I Agree

jessicatarra:
@spydon, It seems like the exit code when running in Windows environments is not reliable, and this causes issues when using some Melos commands.

spydon:
> @spydon, It seems like the exit code when running in Windows environments is not reliable, and this causes issues when using some Melos commands. 

Hmm, there must be some underlying issue for that...

spydon:
Closing this as stale.

PR #789, opened 2024-11-08
fix: Use text script rather than fixed script paths

AlexV525:
## Description

The `melos` executable path could be changed due to environment updates. Using a text script and letting the configuration run in the terminal can reduce the pain of checking out these configurations into version control.

## Type of Change


spydon (approved):

AlexV525:
@spydon Thanks for the approve. I have little time to improve this with some potential tests. If you think it's fine then I will mark this as ready.

spydon:
> @spydon Thanks for the approve. I have little time to improve this with some potential tests. If you think it's fine then I will mark this as ready.

I think it's fine, it looks like it makes sense and I assume you've tested it manually.

spydon (approved):

AlexV525:
Hi @spydon. I'm still seeing the old style of generated configurations. I saw the commit was published in v6.3.2 and I was using the exact same version. Is there something missing from the publish?

spydon:
Hey @AlexV525, everything should be out.
Maybe something went wrong with this PR so that it didn't have any effect then? 🤔

spydon:
Have you tried if it works on the 7.0.0 pre-releasd version?

AlexV525:
I just upgraded my workspace dependency and it seems working. I used to think that a globally activated package would ignore the resolved version in the workspace, but now it appears not.

Issue #792, opened 2024-11-15
fix:  `melos analyze` is running `dart analyze` for Flutter packages


mahmuttaskiran:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Version

6.1.0

### Description

When I run melos analyze, it fetches dependencies for the root project, allowing the analysis to succeed for the root. However, for the packages, it fails because it does not resolve dependencies before running the analysis for each package.

### Steps to reproduce

- Run flutter clean for each package to delete the pubspec.lock files: melos exec --flutter clean.
- Run melos analyze.

### Expected behavior

It should resolve the packages for each package if needed. Like `flutter analyze` do.

mahmuttaskiran:
One more info: It's running `dart analyze` even for flutter workspaces, and that's why it's not resolving dependencies

spydon:
You're always supposed to have run `bootstrap` (which runs `pub get`) before you run any Melos commands.

mahmuttaskiran:
But that's not the case with the flutter analyze command. It resolves dependencies (runs pub get) before analyzing, if needed. We've delivered a fix for this.
https://github.com/invertase/melos/pull/793

PR #793, opened 2024-11-15
fix: Run `flutter analyze` for flutter packages

mnayef95:
## Description
Current `melos analyze` command is running `dart analyze` instead of `flutter analyze` on flutter projects
https://github.com/invertase/melos/issues/792

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/test/commands/analyze_test.dart
+++ b/packages/melos/test/commands/analyze_test.dart
@@ -277,6 +277,34 @@ $ melos analyze
       expect(regex.hasMatch(logger.output.normalizeNewLines()), isTrue);
     });
 
+    test('should run analysis with using flutter', () async {
+      final workspaceDir = await createTemporaryWorkspace();
+
+      await createProject(
+        workspaceDir,
+        PubSpec(
+          name: 'a',
+          dependencies: {
+            'c': HostedReference(VersionConstraint.any),
+            'flutter': const SdkReference('flutter'),
+          },
+        ),
+      );
+
+      final config = await MelosWorkspaceConfig.fromWorkspaceRoot(workspaceDir);
+
+      final melos = Melos(
+        logger: logger,
+        config: config,
+      );
+      await melos.analyze(concurrency: 2);
+
+      final regex =
+          RegExp(r'\$ melos analyze\s+└> flutter analyze --concurrency 2');
+
+      expect(regex.hasMatch(logger.output.normalizeNewLines()), isTrue);
+    });
+
     test('should run analysis with --concurrency 2 flag', () async {
       await melos.analyze(concurrency: 2);
 
--- a/packages/melos/test/commands/analyze_test.dart
+++ b/packages/melos/test/commands/analyze_test.dart
@@ -277,16 +277,25 @@ $ melos analyze
       expect(regex.hasMatch(logger.output.normalizeNewLines()), isTrue);
     });
 
-    test('should run analysis with using flutter', () async {
+    test('should run analysis using flutter & dart', () async {
       final workspaceDir = await createTemporaryWorkspace();
 
       await createProject(
         workspaceDir,
-        PubSpec(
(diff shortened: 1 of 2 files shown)


spydon:
How come you need to run `flutter analyze` instead of `dart analyze`? It should result in the same warnings/errors?

mahmuttaskiran:
> How come you need to run `flutter analyze` instead of `dart analyze`? It should result in the same warnings/errors?

Let me clarify. For Flutter projects using Melos, when neither `melos bootstrap` nor `flutter pub get` is executed for the packages, `dart analyze` will fail. This happens because it does not run `pub get`, and since dependencies are not resolved, `dart analyze` complains about import statements that refer to any unresolved dependencies. (This is just one case; there can be many others, such as referring to a class that belongs to a dependency.)

However, `flutter analyze` ensures that `pub get` is executed before running the analysis, which resolves this issue.

You can reproduce this issue in any Flutter repository using Melos by running `melos clean` and then attempting to analyze the code without bootstrapping the packages.

---
Note: With this fix, do not try to reproduce this issue and understand the problem in the Melos repository itself, as it is not a Flutter workspace and will always use dart analyze, which does not run pub get.

packages/melos/test/commands/analyze_test.dart
```diff
@@ -277,6 +277,34 @@ $ melos analyze
       expect(regex.hasMatch(logger.output.normalizeNewLines()), isTrue);
     });
 
+    test('should run analysis with --concurrency flag', () async {
```
mahmuttaskiran, line 280: Update the desc for the test

mnayef95:
Hello @spydon, I hope you are doing well. According to the [Flutter documentation](https://docs.flutter.dev/reference/flutter-cli), we should use `flutter analyze` instead of `dart analyze`. Additionally, as stated by @mahmuttaskiran, `flutter analyze` will also run `pub get`, making it unnecessary to run bootstrap—especially for large projects with over 100 packages.

Screenshot 2024-11-15 at 5 19 32 PM

spydon:
> Hello @spydon, I hope you are doing well. According to the [Flutter documentation](https://docs.flutter.dev/reference/flutter-cli), we should use `flutter analyze` instead of `dart analyze` for flutter projects. Additionally, as stated by @mahmuttaskiran, `flutter analyze` will also run `pub get`, making it unnecessary to run bootstrap—especially for large projects with over 100 packages.
> 
> Screenshot 2024-11-15 at 5 19 32 PM
> 

If that is the case that flutter analyze is recommended over dart analyze for flutter projects (which isn't actually stated in the docs like you say), we should also follow it the other way around. Dart packages should be using `dart analyze`, but in this PR flutter will be used for all pure dart packages too if there is a flutter package present in the repo.

spydon:
> For Flutter projects using Melos, when neither melos bootstrap nor flutter pub get is executed for the packages, dart analyze will fail.

If you don't run `bootstrap` your monorepo won't be properly set up anyways, so interdependencies in the repo will resolve to pub instead of to their local counterparts, so that `dart analyze` doesn't run `pub get` is not concerning imo.

mnayef95:
> Are your packages not using each other in your repository? If they are and you aren't manually setting path dependencies you won't be running static analysis properly on the actual state of the repo, since it will be pulling down released versions instead. 

Yes, the packages are interdependent, and we are using path dependencies. We tried running `flutter analyze` and `dart analyze`. `dart analyze` fails because some URIs can't be resolved unless `pub get` is run first, so we tried `flutter analyze`, and it worked as it will do pub get and that what we need for static analysis CI step.

I've updated the PR to include the check for Dart packages. I don't think it's an optimal solution, but it's what I could do with the current implementation.

spydon (approved):
Lgtm, thanks for your contribution!

mahmuttaskiran (approved):

PR #799, opened 2024-12-06
fix: Fixed output tests that failed on some terminals

christerswahn:
## Description

Several of the existing test cases failed locally due to inconsistent terminal width logic. This problem surfaced when run on a Mac, using iTerm2 and Bash 5, with a terminal width greater than 80. Some tests, but not all, were hardcoded with output that was 80 wide.

This change makes all the tests consistently use dynamic terminal width in output comparison.

It might be argued that Melos shouldn't write dashed lines longer than 80 anyway, but that's an output design question and beyond the scope of this narrow fix.

Also reinstated two output-checking tests that were skipped since they were sensitive to processing time differences. This was done by simply using regexps for the varying time.

## Type of Change

- [ x] 🛠️ `fix` -- Bug fix (non-breaking change which fixes an issue)
- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/test/commands/analyze_test.dart
+++ b/packages/melos/test/commands/analyze_test.dart
@@ -2,6 +2,7 @@ import 'dart:io';
 
 import 'package:melos/melos.dart';
 import 'package:melos/src/common/io.dart';
+import 'package:melos/src/common/utils.dart';
 import 'package:path/path.dart' as p;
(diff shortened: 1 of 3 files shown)


spydon (approved):
I've always been confused about why it was locked to 80 for the tests to begin with, thanks for the improvement!

PR #954, opened 2025-10-12
feat: add private property for scripts

### LINKED ISSUE #420: request: Add ability to make scripts `private`

### Command

private: true

### Description

Add the ability to mark scripts as `private`.

Not all of the scripts are needed to be exposed by default. As the project grows bigger you will add some scripts that should be used only as a part of aggregated scripts. See that example with a private `cleanup` script:
``` yaml
scripts:
  format:
    description: Format all packages
    run: ...
  analyze:
    description: Analyze all packages
    run: ...
  test:
    description: Test all packages
    run: ...
  cleanup:
    description: Clean up
    private: true  # We do not want to expose this command to the `melos run` output by default
    run: ...
  
  pr_check:
    run: |
      melos run format && \
      melos run analyze && \
      melos run test && \
      melos run cleanup
```

The output for the `melos run` in that case would be:
``` bash
> melos run

1) format
2) analyze
3) test
4) pr_check
```

Introducing the additional parameter `--include-private` would then return the full list of available scripts:
``` bash
> melos run --include-private

1) format
2) analyze
3) test
4) cleanup
5) pr_check
```

### Reasoning

Some of the scripts should be visible by default when running `melos run` command. There could be different reasons why you want to achieve that. Eg:
- a script needed to be only as a part of an aggregated script
- a script needed to be run only during CI/CD pipeline and you want to avoid using the script locally by other developers
- a script makes irreversible changes and it's too dangerous to expose it by default

_No response_

ARau87:
Closes #420 

## Description

- Add the option to mark scripts as `private`
- Avoid that private scripts are included in the prompt and other listings
- Avoid that private scripts are run directly with `run`
- But make it possible to include private scripts as steps of other scripts.
- Add `--include-private` flag to ignore the option 

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/commands/exec.dart
+++ b/packages/melos/lib/src/commands/exec.dart
@@ -129,7 +129,11 @@ mixin _ExecMixin on _Melos {
 
     operation = CancelableOperation.fromFuture(
       pool.forEach<Package, void>(sortedPackages, (package) async {
-        assert(!(failFast && failures.isNotEmpty));
+        if (failFast && failures.isNotEmpty) {
+          packageResults[package.name]?.complete();
+          failures[package.name] = null;
+          return;
+        }
 
         if (orderDependents && !hasImpactfulCycles) {
           final dependenciesResults = await Future.wait(
--- a/packages/melos/lib/src/commands/list.dart
+++ b/packages/melos/lib/src/commands/list.dart
@@ -29,6 +29,7 @@ mixin _ListMixin on _Melos {
         return _listColumn(
           workspace,
           long: long,
+          relativePaths: relativePaths,
         );
       case ListOutputKind.json:
         return _listJson(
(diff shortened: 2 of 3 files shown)


packages/melos/lib/src/commands/run.dart
```diff
@@ -357,8 +365,8 @@ class ScriptNotFoundException implements MelosException {
   @override
   String toString() {
     final builder = StringBuffer(
-      'ScriptNotFoundException: The script $scriptName could not be found in '
-      "the 'pubspec.yaml' file.",
+      'ScriptNotFoundException: A public script $scriptName could not be found '
```
spydon:
```suggestion
      'ScriptNotFoundException: A script named $scriptName could not be found '
```

spydon (approved):
Thanks for your contribution!

Issue #974, opened 2025-12-06
request: Execution with `orderDependents` should be layered


sikrinick:
### Is there an existing feature request for this?

- [x] I have searched the existing issues.

### Description

`_execForAllPackages` with `orderDependents` currently uses topological sorting.  
Concurrent execution is limited only by number of available processors or defined `concurrency` argument.

Unfortunately, this approach doesn't cover cases, when results of a command for a package are used to execute command for package that is next in a topological sorted list.

As an example, those results can files generated by any builder by `build_runner build`.

To support this case, there should be an option to rely on layered topological sort, and have executions paralelized on a layer, but layer execution should be sequential.

For example, instead of
```dart
# commands/exec.dart
...
late final CancelableOperation<void> operation;

operation = CancelableOperation.fromFuture(
  pool.forEach<Package, void>(sortedPackages, (package) async {
...
```
It could be
```dart
# commands/exec.dart
...
for (final packageLayer in sortedPackages) {
  late final CancelableOperation<void> operation;
  
  operation = CancelableOperation.fromFuture(
  pool.forEach<Package, void>(packageLayer, (package) async {
...        
```

### Reasoning

Let's say, we have next project structure:
```
flowchart TD
    app --> common
    app --> module_a
    app --> module_b
    app --> module_c
    app --> module_d
    module_a --> common
    module_b --> common
    module_c --> common
    module_d --> common
    module_a --> module_b
    module_b --> module_c
  ```

<img width="646" height="545" alt="Image" src="https://github.com/user-attachments/assets/1a05b7f1-1ee7-4167-ba60-47f87c1c8512" />

In total — 6 modules.

Reverse topologically sorted list may look next:
`['common', 'module_c', 'module_b', 'module_a', 'module_d', 'app']`

Let's say `Platform.numberOfProcessors` will return 10, so concurrency will be equal to 10.

In this case, an `exec` command will be executed for each of this modules at the same time, as 6 < 10.

Let's say, example project relies on `dart_mappable` in every module and requires to run `build_runner build`.   
If we rename field in `@MappableClass` and run `build_runner build`, we have to try to generate new code for each of its dependents.

Obvious solution would be to use `concurrency: 1`, but that is not scalable for projects, that have large amount of modules.  
For example, in our current project we have 167 modules, and running `build_runner build` sequentially takes around 40 minutes.  
We run it in parallel thanks to `orderDependents` and it decreases up to 10 minutes, so we would genuinely prefer not to limit concurrency.

[...]

PR #976, opened 2025-12-07
feat: Changed from topological sort to layered topologic sort for orderDependents

sikrinick:
## Description

Feature/Fix for https://github.com/invertase/melos/issues/974

1. Refactored `_execForAllPackages` to rely on layered topological sort in case if `orderDependents` is true.
2. Added cyclic dependency check for `exec` in case if `orderDepdents` is true.
3. Removed `dependency failed` checks in case if `failFast: false`
4. Added test for case if `orderDependents: true` with package filter applied.
5. Moved functions that use package `graphs` to `topology.dart` for readability

Generally, this PR is a breaking change, as it changes behaviour of `_execForAllPackages` in case if `orderDependents` is true.
If there are cyclic dependencies — it shouldn't be allowed to run `exec` with `orderDependents`, as in this case we cannot have layers. Thus, with this PR `exec` runs with `orderDependents: true` with cyclic dependencies will not succeed, thus, being a breaking change.
On the other hand, we don't need to maintain `hasImpactfulCycles` failure logic, and we can rely on existing mechanism.
This changes logs behave, so this is another breaking change.

`_PublishMixin` shouldn't be impacted, as check against cyclic dependencies is not needed, due to hard limit of `concurrency: 1` and due to `orderDependents: false`.

On the other hand, there is some topological sort for publish packages.

Breaking changes are visible in changes of test logic.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/common/topology.dart
+++ b/packages/melos/lib/src/common/topology.dart
@@ -1,0 +1,106 @@
+import 'package:cli_util/cli_logging.dart';
+import 'package:collection/collection.dart' hide stronglyConnectedComponents;
+import 'package:graphs/graphs.dart';
+
+import '../package.dart';
+
+/// Sorts [packages] in topological order so they can be published without
+/// errors.
+///
+/// Packages with inter-dependencies cannot be topologically sorted and will
+/// be sorted by name length. This is a heuristic to better handle cyclic
+/// dependencies in federated plugins.
+void sortPackagesForPublishing(List<Package> packages) {
+  final packageNames = packages.map((package) => package.name).toList();
+  final graph = <String, Iterable<String>>{
+    for (final package in packages)
+      package.name: [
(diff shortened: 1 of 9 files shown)


packages/melos/lib/src/common/topology.dart
```diff
@@ -0,0 +1,106 @@
...
+
+/// Filters [packageLayers] to contain only [executablePackages].
+/// Empty layers are dropped.
+List<List<Package>> whereOnlyExecutablePackages(
+  List<List<Package>> packageLayers,
+  List<Package> executablePackages,
+) {
+  final executablePackageNames = {
+    for (final package in executablePackages) package.name,
+  };
```
spydon: Nit, either user this way, or map it like on line 41. Just use the same way on both places.

PR #1010, opened 2026-05-18
feat(scripts): add stdio: inherit for interactive scripts

Prn-Ice:
## Description

Melos pipes a script's stdout/stderr by default so it can prefix child output. As a side effect, the child process has no controlling terminal (its stdin is also unconnected), which breaks any program that requires a real TTY: `tmux attach`, `vim`, `less`, hot-reload key handling in `flutter run` / `dart_frog dev`, etc. They fail with errors like `open terminal failed: not a terminal`.

This adds an opt-in `stdio: inherit` field on script definitions:

    scripts:
      attach:
        run: tmux attach -t my-session
        stdio: inherit

When set, melos uses `ProcessStartMode.inheritStdio` so the child's stdin/stdout/stderr are wired directly to melos's parent terminal. The trade-off is documented on the enum: melos drops its indented output prefix for the duration of that script in exchange for the TTY.

The default remains `pipe`; existing scripts are unaffected.

Validation rejects two combinations that cannot work:
  - `stdio: inherit` + `exec` (cannot multiplex one TTY across N parallel package children)
  - `stdio: inherit` + `steps` (`steps` run inside a persistent shell; put the inherit flag on the individual scripts referenced from `steps` instead)

Changes:
  - scripts.dart: new ProcessStdio enum, Script.stdio field, YAML parser, validation, equality/hashCode/toString/toJson coverage.
  - common/utils.dart: startCommandRaw accepts `mode`; startCommand accepts `inheritStdio` and branches before subscribing to process.stdout/stderr (those streams are unavailable in inheritStdio mode and would throw a StateError if listened to).
  - commands/run.dart: passes `script.stdio == ProcessStdio.inherit` through to startCommand.
  - melos.dart: exports the new ProcessStdio enum.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/scripts.dart
+++ b/packages/melos/lib/src/scripts.dart
@@ -102,35 +102,28 @@ enum ProcessStdio {
   /// Child stdout and stderr are piped through melos so it can prefix output
   /// with the script's indent. Child stdin is not connected. This is the
   /// default and is what every script gets without an explicit `stdio` key.
-  pipe('pipe'),
+  pipe,
 
   /// Child stdin, stdout, and stderr are inherited from melos's own terminal,
   /// giving the script a real TTY. Required for any program that needs an
   /// attached terminal — `tmux attach`, `vim`, `less`, `flutter run` /
   /// `dart_frog dev` hot-reload keys. Melos drops its indented output prefix
   /// for the duration of the script in exchange for the TTY.
-  inherit('inherit')
-  ;
-
-  const ProcessStdio(this.yamlValue);
-
-  /// The string used to select this mode in melos config (e.g.
-  /// `stdio: inherit`).
-  final String yamlValue;
+  inherit;
 
   /// Parses the value of a script's `stdio:` config key.
   ///
   /// Throws [MelosConfigException] if [value] is not a recognised mode.
-  static ProcessStdio fromYaml(
+  static ProcessStdio fromString(
     String value, {
     required String scriptName,
   }) {
     for (final mode in ProcessStdio.values) {
-      if (mode.yamlValue == value) {
+      if (mode.name == value) {
         return mode;
       }
     }
-    final allowed = ProcessStdio.values.map((m) => m.yamlValue).join(', ');
+    final allowed = ProcessStdio.values.map((m) => m.name).join(', ');
     throw MelosConfigException(
       'Invalid value "$value" for "stdio" on script $scriptName. '
       'Expected one of: $allowed.',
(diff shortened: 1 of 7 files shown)


packages/melos/lib/src/scripts.dart
```diff
@@ -96,6 +96,48 @@ ExecOptions(
...
+  /// `dart_frog dev` hot-reload keys. Melos drops its indented output prefix
+  /// for the duration of the script in exchange for the TTY.
+  inherit('inherit')
+  ;
+
+  const ProcessStdio(this.yamlValue);
+
+  /// The string used to select this mode in melos config (e.g.
+  /// `stdio: inherit`).
+  final String yamlValue;
```
spydon: This is not needed, since it's the same as `ProccessStdio.name`

packages/melos/lib/src/scripts.dart
```diff
@@ -96,6 +96,48 @@ ExecOptions(
...
+
+  /// Parses the value of a script's `stdio:` config key.
+  ///
+  /// Throws [MelosConfigException] if [value] is not a recognised mode.
+  static ProcessStdio fromYaml(
```
spydon:
```suggestion
  static ProcessStdio fromString(
```

packages/melos/lib/src/scripts.dart
```diff
@@ -96,6 +96,48 @@ ExecOptions(
...
+    String value, {
+    required String scriptName,
+  }) {
+    for (final mode in ProcessStdio.values) {
+      if (mode.yamlValue == value) {
```
spydon:
```suggestion
      if (mode.name == value) {
```

packages/melos/lib/src/scripts.dart
```diff
@@ -252,6 +295,17 @@ class Script {
           )
         : [];
 
+    final stdioYaml = yaml['stdio'];
```
spydon: Since we don''t need to check is it is a list or such here, do something like this:
```dart
final stdioValue = assertKeyIsA<String?>(
    key: 'stdio',
    map: yaml,
    path: scriptPath,
  );
  final stdio = stdioValue != null
      ? ProcessStdio.fromYaml(stdioValue, scriptName: name)
      : ProcessStdio.pipe;
```

spydon (commented):
Some small comments, but other than that it looks good!
If you can sign the CLA and fix those comments we can get this merged. :)

Prn-Ice:
Heads up — pushed one extra drive-by commit (7a97a18) to unblock CI.

The CI run hit:

> `/home/runner/work/melos/melos/packages/melos/lib/src/command_runner.dart:94:2 The annotation 'override' can only be used on fields, getters, methods, or setters.`

That's `@override` applied to the top-level `melosEntryPoint` function, which has been on `main` for a while but a stricter analyzer now refuses it. I removed the annotation in `command_runner.dart` so the analyze step would pass.

Happy to pull that into a separate PR if you'd prefer to keep this one focused on `stdio: inherit` — let me know.

CLA: I'll sign now.

spydon:
> Happy to pull that into a separate PR if you'd prefer to keep this one focused on stdio: inherit — let me know.

It's all good, since it's such a small thing we can bundle it into this one. Thanks for fixing that.

spydon (approved):
Thanks for your contribution

Prn-Ice:
> Thanks for your contribution

Thanks @spydon  😁

spydon:
@Prn-Ice the code looks good, but I just realized that docs are missing, can you please add those? 
The docs live in the root [docs](https://github.com/invertase/melos/tree/main/docs) directory.

Prn-Ice:
Added a `## stdio` section to `docs/configuration/scripts.mdx` (commit in the latest push) — covers `pipe` (default) vs `inherit`, when to reach for `inherit`, and that it can't be combined with `exec`/`steps`.

I signed the cla already. Not sure why it's still failing

Prn-Ice:
Ohh.
I committed with another account by accident.

spydon:
@Prn-Ice  This is now released in v7.8.0 :)

Issue #924, opened 2025-08-19
fix: ERROR: Could not find a command named "analyze"


mockturtl:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Version

7.0.0

### Description

After upgrading to 7.0.0, `melos analyze`  gives

```
Could not find a command named "analyze".
```

Was `analyze` removed?  https://melos.invertase.dev/~melos-latest/commands/analyze

It doesn't appear in the command list (`melos --help`), but there's no mention in the CHANGELOG.

### Steps to reproduce

 `melos analyze`

### Expected behavior

It should run `flutter analyze` in each package (they may have different `analysis_options.yaml` config files).

### Other

- [ ] I'm interested in working on a PR for this.

spydon:
Ah, it seems like we forgot to update the docs regarding `melos analyze`. When running `dart analyze` or `flutter analyze` in a pub workspace it runs it in the whole monorepo, so `melos analyze` is no longer needed.

xaethos:
> When running dart analyze or flutter analyze in a pub workspace it runs it in the whole monorepo, so melos analyze is no longer needed.

@spydon from my testing, I believe this is not entirely correct, or at least it does not address the issue. `dart analyze` runs the analysis in the current directory, not the whole workspace.

Switching to the root directory is trivial, but then comes the real issue: now `dart analyze` runs _in the whole monorepo_, regardless if it is part of the workspace. In my case our monorepo includes a large amount of non-Dart sources, and an `experimental` directory with packages not included in the workspace (with many analysis errors).

I can exclude `experimental` in `analysis_options.yaml`, but I don't want to maintain the list of non-Dart excludes. The analysis options file does not seem to support _path_ includes, but even if it did, it would just be a repeat of the `workspaces` list.

spydon:
Ah, if that is the case I'll reopen this issue.

I thought since the analyzer doesn't fork itself when running in a pub workspace I assumed that it always ran on the whole workspace, but that might not be the case then. If that's true I'm open to bringing `melos analyze` back.
I won't have much bandwidth doing that in the coming weeks, but PRs are welcome and I can review and publish a new version.

rumitvn:
Opened #1032 to fix the broken markdown in the migration note: the `### 'melos analyze` heading opened with a stray quote and rendered as literal text. Scoped to just the heading/punctuation fix, leaving the technical discussion about `dart analyze` workspace coverage untouched.

[...]

PR #1032, opened 2026-06-14
docs: fix broken markdown heading for removed `melos analyze`

rumitvn:
Addresses #924

The v7 migration note about the removed `melos analyze` command has a broken heading:

```
### 'melos analyze` no longer exists
```

It opens the inline-code span with a stray straight quote (`'`) and closes it with a backtick, so it renders as literal `'melos analyze` instead of code. This fixes the heading to use backticks on both sides (matching how `melos analyze` is written in the body just below), and ends the following sentence with a period instead of a trailing comma.

Scope is intentionally limited to the rendering/punctuation fix; it doesn't touch the technical wording about whether `dart analyze` covers the full workspace (still under discussion in #924).

### DIFF
--- a/docs/guides/migrations.mdx
+++ b/docs/guides/migrations.mdx
@@ -93,10 +93,10 @@ If you are using an older version of the Dart SDK, you can still use Melos
 with pub workspaces, but then you have to rely on the `7.0.0-dev.x` versions.
 </Info>
 
-### 'melos analyze` no longer exists
+### `melos analyze` no longer exists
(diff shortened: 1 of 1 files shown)


spydon (approved):

spydon:
Thanks for your contribution @rumitvn, can you sign the CLA so that we can merge it?

rumitvn:
Sure thing. Please check again @spydon. Love your nice work

Issue #815, opened 2025-01-05
request: support `--no-pub`


lishaduck:
### Is there an existing feature request for this?

- [X] I have searched the existing issues.

### Description

The Flutter and Dart CLIs offer an option to turn off calling `pub get` before they run, which slows them down.
It's called `--no-pub`, and can be passed to analyze, test (although Melos doesn't have a test command), etc.

### Reasoning

I'm on a rather limited machine, so reducing redundant work is important.
`pub get` can take up to a minute, and while pub workspaces will unify most stuff, they'll also (probably) slow down `pub get` even more.

### Additional context and comments

Alternatively, passing unknown flags to the underlying command might work as well.

spydon:
Makes sense, would you be interested in working on that issue?

lishaduck:
> Makes sense, would you be interested in working on that issue?

Interested, yes; actually do it, probably someday? 😁
Now that I think about it, it'd probably be a good introduction before I tackle #747 (I plan on getting to that if no one else does, but I'm pretty busy with other obligations at the moment).

FranRiadigos:
Hi,

As of version `7.0.0-dev.9`, `melos bootstrap` runs `pub get` before propagating shared dependencies.
This makes a contradiction with that functionality. 
Should propagate first and then run `pub get`.

For example:
In our monorepo we have a dep that, say requires `meta` version 2, while most of our packages use `meta` 1.
Once you switch the dep with the `meta` version 2 requirement, since the bootstrap runs `pub get` earlier, it simply crashes with a dep conflict, regardless if you update the shared `meta` dep, since shared deps never get to the propagation process.
(So one ends up manually replacing everywhere, hence not strictly a blocker)

I saw on https://github.com/invertase/melos/pull/816 it's been added an `--offline` mode
That's actually not `--no-pub`, and I was hoping to see whether this was implemented to kind a sort this issue out

Since all of us are generally busy, not sure if I can dig into fixing this earlier pub get issue, but I consider it a bug.

Please, let me know if I'm wrong or if there's any quick workaround apart from manually replacing

@spydon  @exaby73

Thanks,
Fran

spydon:
>  Should propagate first and then run pub get.

Unfortunately it can't, because the initial `pub get` gets the Melos version that is going to be used for the execution of the command.

>  In our monorepo we have a dep that, say requires meta version 2, while most of our packages use meta 1.

[...]

PR #1038, opened 2026-07-03
feat(melos): add melos test command

### LINKED ISSUE #830: request:  melos test

### Command

melos test

### Description

Pub Workspaces do not support running `dart test` or `flutter test` across all packages in the workspace. Developers still need to manually navigate to each package or script custom solutions to run tests across the entire monorepo.

### Reasoning

1. One command to run all tests across the workspace, regardless of package type
2. Eliminate the need to create a melos script that runs the tests

I know it's already possible to create a custom Melos script to achieve this, but it would be helpful to have a built-in solution, similar to how `melos analyze` automatically runs `dart analyze` for Dart projects and `flutter analyze` for Flutter projects.

Gurleen-kansray:
## Description

Adds a built-in `melos test` command that runs `dart test` (or `flutter test` for Flutter packages) across every workspace package that has a `test/` directory, in a single invocation. Mirrors `melos analyze`'s package-type detection to choose the right tool per package, and reuses the existing `--concurrency` and package-filtering conventions from `melos exec`/`melos format`.

Closes #830.

### Design notes

- Not replicating `very_good_cli`'s single-isolate batching for v1 — straightforward per-package invocation, as discussed on the issue. Happy to revisit if that's wanted.
- `dart test`/`flutter test` scope to the package's own `test/` directory, so this command doesn't depend on the workspace walk-up scoping work being done separately for `melos analyze`.

### Verification

The tracking issue raised a concern about whether package-type detection (`isFlutterPackage`) works correctly under pub workspace resolution. Checked this directly rather than assuming: `isFlutterPackage` is based purely on each package's own declared `dependencies:` in its `pubspec.yaml` (`pubspec.dependencies.keys`), which is unaffected by how pub workspaces resolve versions. Added a regression test (`correctly detects a flutter package inside a pub workspace`) confirming this holds under actual `resolution: workspace` mode.

Also ran:
- `dart analyze` — clean
- `dart format --set-exit-if-changed` — clean
- `dart test test/commands/test_test.dart` — 4/4 passing (skips packages without `test/`, reports when none match, propagates a real failing test's exit code, and the workspace/flutter-detection regression check)

### Known open question

The repo's own `pubspec.yaml` has a custom `melos.scripts.test` entry that runs the dev suite. Melos's built-in-command registration skips a built-in command if a custom script of the same name exists — so as it stands, running `melos test` in *this* repo will still invoke the legacy script, not this new command. Left `pubspec.yaml` untouched rather than deciding this unilaterally — happy to remove/rename the old script in a follow-up commit here if that's the preferred resolution, just wanted a maintainer call on it first.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/test/command_runner_test.dart
+++ b/packages/melos/test/command_runner_test.dart
@@ -59,5 +59,29 @@ void main() {
       // The command should be hidden (custom script behavior)
       expect(command!.hidden, isTrue);
     });
+    test('custom test script overrides built-in test command', () async {
+      final workspaceDir = await createTemporaryWorkspace(
+        configBuilder: (path) => MelosWorkspaceConfig(
+          path: path,
+          name: 'test_package',
+          packages: [
+            createGlob('packages/**', currentDirectoryPath: path),
+          ],
+          scripts: const Scripts({
+            'test': Script(name: 'test', run: 'echo "custom test"'),
+          }),
+        ),
+        workspacePackages: [],
+      );
+
+      final config = await MelosWorkspaceConfig.fromWorkspaceRoot(workspaceDir);
+      final runner = MelosCommandRunner(config);
+
+      final command = runner.commands['test'];
+      expect(command, isNotNull);
+      // The command should be hidden (custom script behavior), confirming
+      // the built-in TestCommand was NOT registered in its place.
+      expect(command!.hidden, isTrue);
+    });
   });
 }
--- a/packages/melos/test/command_runner_test.dart
+++ b/packages/melos/test/command_runner_test.dart
@@ -23,7 +23,9 @@ void main() {
         workspacePackages: [],
       );
 
-      final config = await MelosWorkspaceConfig.fromWorkspaceRoot(workspaceDir);
+      final config = await MelosWorkspaceConfig.fromWorkspaceRoot(
+        workspaceDir,
+      );
       final runner = MelosCommandRunner(config);
 
       expect(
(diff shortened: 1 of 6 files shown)


spydon:
I think #815 is wrongly linked, you were probably thinking of some other issue I guess?

spydon (commented):
Looks good! Can you just add a test to make sure that if the user has their own `test` script defined that that takes precedence?

Gurleen-kansray:
You're right, that was my mistake ....#815 was actually referenced in the original issue in relation to `--no-pub` defaulting, not the single-isolate batching point. There isn't a specific tracked issue for the batching idea itself that I'm aware of. I'll drop the incorrect reference.

packages/melos/test/command_runner_test.dart
```diff
@@ -59,5 +59,29 @@ void main() {
...
+          }),
+        ),
+        workspacePackages: [],
+      );
+
+      final config = await MelosWorkspaceConfig.fromWorkspaceRoot(workspaceDir);
+      final runner = MelosCommandRunner(config);
+
+      final command = runner.commands['test'];
+      expect(command, isNotNull);
```
spydon, line 90: Does this really verify it? I think it would be a more solid test to run it and check the output.

Gurleen-kansray:
Thanks @spydon, done!

spydon (approved):

PR #1041, opened 2026-07-03
fix(version): handle quoted range constraints in dependent rewrites

### LINKED ISSUE #1039: melos version corrupts quoted space-separated range constraints in dependents (e.g. '">=0.4.0 <1.0.0"' becomes '^0.6.0 <1.0.0"')
## Summary
When `melos version` bumps a package and rewrites dependents' constraints, a dependency declared with a quoted, space-separated range constraint gets **corrupted** instead of rewritten.

## Reproduction
Workspace package `jose_plus` depends on sibling `crypto_keys_plus`:
```yaml
dependencies:
  crypto_keys_plus: ">=0.4.0 <1.0.0"
```
Run `melos version --yes --no-git-tag-version` with a versionable (feat/fix) commit touching `crypto_keys_plus`. Result in the dependent's pubspec:
```yaml
  crypto_keys_plus: ^0.6.0 <1.0.0"
```
which then fails any subsequent pub command:
```
Error on line 14, column 21 of packages/jose_plus/pubspec.yaml: Invalid version constraint: Cannot include other constraints with "^" constraint in "^0.6.0 <1.0.0"".
```
Real-world occurrence: https://github.com/Bdaya-Dev/oidc/actions/runs/28638001109 (melos 7.8.2 via bluefireteam/melos-action). Reproduced identically on 7.8.2 from a clean clone.

## Root cause
`packages/melos/lib/src/package.dart` lines 75-76 (identical on tags `melos-v7.8.2`, `melos-v8.0.0`, and `main`):
```dart
const _versionConstraintRegExp =
    r"""(?<version>any|["'^<>=]*\d+\.\d+\.\d+['"<>=\w\-.+_]*)""";
```
The trailing character class contains **no whitespace**, so against `">=0.4.0 <1.0.0"` the match stops at the space, capturing only `">=0.4.0`. The rewrite in `packages/melos/lib/src/commands/version.dart` (~639-644, with the unconditional caret rewrite from ~541-547) then splices the new `^x.y.z` over that partial match, leaving the tail ` <1.0.0"` behind as literal garbage.

## Expected
Either rewrite the full range constraint correctly, or leave constraints it cannot parse untouched (with a warning) — anything but writing an unparseable pubspec.

## Environment
melos 7.8.2 (also verified regex unchanged on 8.0.0 and current main); Dart/Flutter stable 3.44.x; pub workspace monorepo.

## Workaround
Normalize internal dependency constraints to plain caret form — melos's rewrite handles those correctly (and its caret rewrite would collapse wide ranges on the next bump anyway).

ahmednfwela:
Fixes #1039

## Summary

`melos version`'s dependent-constraint rewrite used a regex (`_versionConstraintRegExp` in `packages/melos/lib/src/package.dart`) whose trailing character class had no whitespace. Against a quoted, space-separated range constraint like `">=0.4.0 <1.0.0"`, the regex only matched up to the first space (`">=0.4.0`), and the caret rewrite in `commands/version.dart` was then spliced over that partial match — leaving the rest of the old constraint (` <1.0.0"`) behind as literal garbage:

```yaml
# before
crypto_keys_plus: ">=0.4.0 <1.0.0"
# after (broken)
crypto_keys_plus: ^0.6.0 <1.0.0"
```

This produces an unparseable `pubspec.yaml`:

```
Invalid version constraint: Cannot include other constraints with "^" constraint in "^0.6.0 <1.0.0"".
```

Rather than patching the regex, this PR migrates the plain-dependency, hosted-dependency (`hosted: {..., version: ...}`), and git-tag-pattern-dependency (`git: {..., tag_pattern: ...}` + sibling `version:`) rewrite paths in `_setDependencyVersionForPackage` to `YamlEditor` — the same tool melos already uses a few lines above for the package's own `version:` bump (`_setPubspecVersionForPackage`), and in `bootstrap.dart`/`init.dart`. Targeting the constraint by its actual YAML path instead of splicing text eliminates the whole bug class (quoting, embedded spaces, trailing comments) rather than just the one reported shape, and four now-dead regexes (`_versionConstraintRegExp`, `dependencyVersionReplaceRegex`, `hostedDependencyVersionReplaceRegex`, `gitTagPatternDependencyVersionReplaceRegex`) are removed.

The git `ref:` tag rewrite (`dependencyTagReplaceRegex`) is left as-is: its version is fused into a single tag string (`name-vX.Y.Z`) rather than being its own scalar node, so it isn't a good fit for a path-based `YamlEditor` update, and it was never affected by this bug (tag names can't contain quotes or spaces).

**Note on scope (`dependency_overrides`)**: the old regex-based rewrite scanned the whole pubspec text with `replaceAllMapped`, so if a dependency of the same name also appeared under `dependency_overrides` with a plain, version-like scalar value, it would be rewritten there too, regardless of intent. The new plain- and hosted-dependency rewrites are scoped precisely to the `dependencies`/`dev_dependencies` section `package.pubspec` actually resolved the dependency from, and no longer reach into `dependency_overrides`. This is a deliberate, and arguably more correct, behavior change — a local override generally shouldn't be silently bumped by `melos version` — and no existing test relies on the old cross-section sweep. The git-tag-pattern path is unaffected: it already tracks its own section (which can be `dependency_overrides`) via `GitTagPatternDependency.section`.

## Test evidence

**Unit/integration tests** — new regression coverage in `test/commands/version_test.dart` under `dependent constraint rewrite (regression for #1039)` exercises the real `melos version` command end-to-end (not just the rewrite function in isolation) for: the exact bug case (double- and single-quoted space-separated ranges), unquoted caret, `any`, an exact pin, a prerelease+build-metadata suffix, a trailing comment, an untouched sibling dependency, and the hosted- and git-tag-pattern-dependency variants of the same quoted-range bug. All pass with the fix; `dart test test/commands/version_test.dart test/package_test.dart` and `dart analyze`/`dart format --set-exit-if-changed` are clean.

**Bug reproduced against unpatched code** — running the new bug-case test (`quoted, space-separated range constraint (exact bug from #1039)`) against the original, unmodified `package.dart`/`version.dart` fails with exactly the corruption from the issue:

```
Expected: return normally
  Actual: <Closure: () => Pubspec>
   Which: threw ParsedYamlException:<ParsedYamlException: line 7, column 6: Unsupported value for "a". Cannot include other constraints with "^" constraint in "^0.6.0 <1.0.0"".
            ,
          7 |   a: ^0.6.0 <1.0.0"
            |      ^^^^^^^^^^^^^^
            '>
```

**End-to-end against the real repro** — cloned [Bdaya-Dev/oidc](https://github.com/Bdaya-Dev/oidc) at `d60f050` (the commit from the issue, before the constraints were manually normalized upstream), added a `dependency_overrides: melos: {path: ...}` pointing at this branch's local checkout, made a `fix(crypto_keys_plus):` commit, and ran the exact failing sequence:

```
dart run melos version --yes --no-git-tag-version --no-git-commit-version --scope="crypto_keys_plus,jose_plus"
```

Before: `packages/jose_plus/pubspec.yaml` had `crypto_keys_plus: ">=0.4.0 <1.0.0"`. After: `crypto_keys_plus: ^0.5.1` — clean, with `x509_plus`'s own (untouched, unrelated) quoted range constraint left exactly as-is. `flutter pub get` on the resulting workspace succeeds (`Got dependencies!`), confirming the rewritten pubspec is valid, in contrast to the `ParsedYamlException` unpatched melos produces on the same input.

**Full suite parity** — ran `dart test` for `packages/melos` (~400 tests) twice on the same machine: once on this branch, once on an untouched clone of `main`. Diffing the two failure sets directly (not relying on `dart test`'s own truncated summary) shows 24 failures common to both runs, all sharing the same signatures unrelated to this change (a broken terminal handle under the sandbox this was run in, crashing interactive prompts regardless of code; and antivirus/file-lock contention on temp-dir cleanup, which cascades into unrelated tests). The small number of additional failures unique to each run individually share those same signatures, land on tests outside `package_test.dart`/the constraint-rewrite paths, and vary between runs - consistent with non-deterministic environment noise, not a regression. Every failure touching dependency-constraint rewriting passes on this branch in every run.

## Type of Change

### DIFF
--- a/packages/melos/lib/src/commands/version.dart
+++ b/packages/melos/lib/src/commands/version.dart
@@ -615,41 +615,56 @@ mixin _VersionMixin on _RunMixin {
     final isGitDependency = dependency is GitDependency;
     final isGitTagDependency = dependency is GitTagPatternDependency;
 
-    var updatedContents = pubspecContent;
-    if (isExternalHostedDependency) {
-      updatedContents = pubspecContent.replaceAllMapped(
-        hostedDependencyVersionReplaceRegex(dependencyName),
-        (match) => '${match.group(1)}$dependencyVersion',
+    // The section (`dependencies` or `dev_dependencies`) that
+    // `package.pubspec` actually resolved [dependency] from. Git tag pattern
+    // dependencies track their own section separately, since they can also
+    // live under `dependency_overrides` (see [GitTagPatternDependency]).
+    final section = normalDependency != null
+        ? 'dependencies'
+        : 'dev_dependencies';
+
+    String? updatedContents;
+    if (isGitTagDependency) {
+      updatedContents = _rewriteDependencyVersionAtPath(
+        pubspecContent: pubspecContent,
+        path: [dependency.section, dependencyName, 'version'],
+        dependencyVersion: dependencyVersion,
+        // Git tag pattern dependencies preserve whether the *existing*
+        // constraint used caret syntax, rather than always following
+        // [dependencyVersion]'s own style.
+        preserveCaretStyle: true,
       );
-    } else if (isGitTagDependency) {
-      updatedContents = pubspecContent.replaceAllMapped(
-        gitTagPatternDependencyVersionReplaceRegex(dependencyName),
-        (match) {
-          final hasCaret = match.group(2) == '^';
-          final prefix = match.group(1)!;
-          final replacement =
-              '$prefix${hasCaret ? '^' : ''}'
-              '${dependencyVersion.toString().replaceFirst('^', '')}';
-          return replacement;
-        },
+    } else if (isExternalHostedDependency) {
+      updatedContents = _rewriteDependencyVersionAtPath(
+        pubspecContent: pubspecContent,
+        path: [section, dependencyName, 'version'],
+        dependencyVersion: dependencyVersion,
+        preserveCaretStyle: false,
       );
     } else if (isGitDependency &&
         workspace.config.commands.version.updateGitTagRefs) {
-      updatedContents = pubspecContent.replaceAllMapped(
+      // The version here is fused into a single git tag string (e.g.
+      // `foo-v1.2.3`) rather than being its own scalar node, so this isn't a
+      // good fit for a `YamlEditor`-based rewrite; it's also not affected by
+      // the quoted-range-constraint bug the other branches fix, since tag
+      // names never contain quotes or spaces.
+      final rewritten = pubspecContent.replaceAllMapped(
         dependencyTagReplaceRegex(dependencyName),
         (match) =>
             '${match.group(1)}$dependencyName-'
             'v${dependencyVersion.min ?? dependencyVersion.max!}',
       );
+      updatedContents = rewritten == pubspecContent ? null : rewritten;
     } else {
-      updatedContents = pubspecContent.replaceAllMapped(
-        dependencyVersionReplaceRegex(dependencyName),
-        (match) => '${match.group(1)}$dependencyVersion',
+      updatedContents = _rewriteDependencyVersionAtPath(
+        pubspecContent: pubspecContent,
+        path: [section, dependencyName],
+        dependencyVersion: dependencyVersion,
+        preserveCaretStyle: false,
       );
     }
 
-    // Sanity check that contents actually changed.
-    if (pubspecContent == updatedContents) {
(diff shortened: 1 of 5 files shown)


spydon:
@ahmednfwela can you sign the CLA and rebase on main? :)

spydon (approved):

PR #1058, opened 2026-08-13
fix(version): honour `--yes` for manual versioning prompts

### LINKED ISSUE #1057: Regression: --yes flag ignored in melos version -V (manual versioning), throws StdinException in non-interactive environments

### Version

8.2.2

### Description

The `--yes` flag is not respected when using `melos version` with manual versioning (`-V` / `--manual-version`), causing a `StdinException` crash in non-interactive environments (no TTY).

This is a regression of the fix from #66, which resolved the same issue in melos 0.4.6.

### Steps to reproduce

In any environment without an attached TTY (CI, Claude Code, piped shell, etc.), run:

```bash
dart run melos version --yes -V my_package:patch
```

### Expected behavior

The `--yes` flag skips all confirmation prompts and the command completes successfully, as documented and as was the case before.

### Actual behavior

The command crashes before reading any input:

```
melos version
  └> /path/to/workspace

Found commits for manually versioned package "my_package".
Unhandled exception:
StdinException: Error getting terminal echo mode, OS Error: Operation not supported by device, errno = 19
#0      Stdin.echoMode (dart:io-patch/stdio_patch.dart:84:7)
#1      get (package:prompts/prompts.dart:60:27)
#2      getBool (package:prompts/prompts.dart:176:16)
#3      promptBool.<anonymous closure> (package:melos/src/common/utils.dart:251:19)
#4      _promptWithTerminal (package:melos/src/common/utils.dart:204:21)
#5      promptBool (package:melos/src/common/utils.dart:250:10)
#6      _VersionMixin._version.<anonymous closure> (package:melos/src/commands/version.dart:303:34)
```

The `prompts` library tries to access `stdin.echoMode` (to set up interactive input) **before** the `--yes` flag is checked, so it crashes immediately in non-TTY contexts. This makes `melos version -V` completely unusable in CI or any scripted environment.

Note: `melos publish --yes --no-dry-run` does **not** have this problem — the `--yes` flag works correctly there.

### Environment

- melos: 8.2.2
- Dart SDK: 3.12.2 (stable) on macos_arm64
- OS: macOS (Darwin)

### Additional context

- Original fix (melos 0.4.6): #66
- The issue can be worked around by piping `y` via `expect` or manually editing `pubspec.yaml` / `CHANGELOG.md` and creating git tags by hand.

Lyokone:
Fixes #1057 — "Regression: `--yes` flag ignored in `melos version -V` (manual versioning), throws StdinException in non-interactive environments".

## Root cause

Two independent defects, both needed for the reported crash:

1. **`--yes` ignored (primary).** In `packages/melos/lib/src/commands/version.dart`, the manual-versioning changelog block called `promptBool(...)` (and then `promptInput(...)`) without consulting `force` — unlike the "shall I continue?" prompt further down, which correctly does `force || promptBool()`. `--yes` maps to `force`, so manual versioning prompted regardless.
2. **Prompt fallback never engaged (secondary).** `_promptWithTerminal` in `packages/melos/lib/src/common/utils.dart` only fell back to non-interactive defaults when `!stdin.hasTerminal`. In the reporter's environment `stdin.hasTerminal` is `true`, but `package:prompts` touching `stdin.echoMode` throws `StdinException` ("Operation not supported by device, errno = 19"), so melos crashed instead of degrading.

## Changes

- `version.dart`: gate `promptForMessage` and the follow-up `promptInput` on `!force`, preserving the existing no-prompt defaults (`defaultsToWithoutPrompt: false`, `defaultUserChangelogMessage`). With `--yes` the behaviour is now identical to running in a non-interactive shell.
- `utils.dart`: wrap `runPrompt()` in a `try`/`catch` on `StdinException` and fall through to the non-interactive path below it, so a terminal that claims to exist but does not work degrades rather than crashing.

## Tests

Two regression tests, each verified to **fail before** the corresponding fix and **pass after**:

- `test/commands/version_test.dart` — runs `melos.version(force: true, manualVersions: ...)` inside an `IOOverrides` zone with a `Stdin` that reports `hasTerminal == true` and throws on every operation, so any prompt attempt fails the test. Asserts the version bump and the default changelog entry are still written.
- `test/utils_test.dart` — `promptBool`/`promptInput` fall back to their defaults when the terminal throws `StdinException`, and still throw `PromptException` when `requirePrompt: true`.

Verified with Dart 3.12.2: `dart analyze packages/melos` clean, `dart format` clean, and `dart test test/commands/version_test.dart test/utils_test.dart` (from `packages/melos`) fully green (36 tests).

## Notes for review

- No CHANGELOG edit — melos generates changelogs via `melos version`.
- The full test suite beyond these two files was not run; the two affected suites pass in full.

### DIFF
--- a/packages/melos/lib/src/commands/version.dart
+++ b/packages/melos/lib/src/commands/version.dart
@@ -300,19 +300,26 @@ mixin _VersionMixin on _RunMixin {
                 'Found commits for manually versioned package "$name".',
               );
 
-              promptForMessage = promptBool(
-                message:
-                    'Do you want to provide an additional changelog entry '
-                    'message?',
-                defaultsToWithoutPrompt: false,
-              );
+              // `--yes`/`force` skips all confirmation prompts, so never
+              // prompt here; fall back to the same default that is used when
+              // no prompt is possible.
+              promptForMessage =
+                  !force &&
+                  promptBool(
+                    message:
+                        'Do you want to provide an additional changelog entry '
+                        'message?',
+                    defaultsToWithoutPrompt: false,
+                  );
             }
 
             if (promptForMessage) {
-              userChangelogMessage = promptInput(
-                'Provide a changelog entry message',
-                defaultsTo: defaultUserChangelogMessage,
-              );
+              userChangelogMessage = force
+                  ? defaultUserChangelogMessage
+                  : promptInput(
+                      'Provide a changelog entry message',
+                      defaultsTo: defaultUserChangelogMessage,
+                    );
             }
           }
 
--- a/packages/melos/lib/src/common/utils.dart
+++ b/packages/melos/lib/src/common/utils.dart
@@ -201,7 +201,14 @@ T _promptWithTerminal<T>(
   bool requirePrompt = false,
 }) {
   if (stdin.hasTerminal) {
-    return runPrompt();
+    try {
+      return runPrompt();
+    } on StdinException {
+      // `stdin.hasTerminal` can be true while stdin does not actually support
+      // the terminal operations the prompt requires (e.g. reading the echo
+      // mode fails with errno 19 in some sandboxed/piped environments).
+      // Degrade to the non-interactive behaviour below instead of crashing.
+    }
   }
 
   if (!requirePrompt) {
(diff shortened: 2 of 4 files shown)


spydon (approved):
Lgtm, thanks for your contribution

spydon:
This is now released in Melos v8.3.0

PR #1061, opened 2026-08-15
fix(publish): only report success when dart pub exits 0

### LINKED ISSUE #966: fix:  Inconsistent exit code / output with melos exec dart pub publish --dry-run when publishing older patch version

### Version

7.0.0-dev.8

### Description

I have a monorepo managed by Melos.
One of the packages (X) is currently at version 9.0.0 on the registry.
I’m trying to publish a backport patch 8.5.1 (while 8.5.0 already exists).

When I run:
```shell
melos publish --dry-run --yes
```

The output shows:

`A FAILED tag for the package (exit code 65)`
But then “All packages were validated successfully”
And Melos process exits with code 0 (echo %ERRORLEVEL% or check $LASTEXITCODE on PowerShell)

Image

I use it in my CI and my pipeline fails even though the dry run is successful.

However if I run the actual publish (no --dry-run), it works fine and publishes 8.5.1 successfully.

So the only problem is the DX and reporting of the dry-run mode:
- It reports a failure even though the version is valid
- It returns an exit code 0 even though it prints “FAILED”

### Steps to reproduce

- Create a monorepo package and release a major version (ex: 2.0.0)
- Create a backport release from a previous major (1.0.0)
- Run:
```shell
melos publish --dry-run --yes
```

### Expected behavior

Dry-run should either :
- Exit with code 0 if sucess
- Exit with code >0 but do not display a success message

### Screenshots

_No response_

When using dart command in the package (not the monorepo), It has the same behavior, the exit code is > 0 in dry-run, but when running the actual publish, it does work and exit code is 0. So the inconsistency between dry-run and actual publish is on dart's end. The point of this issue is simply the inconsistency between the two messages “FAILED” and “SUCCESS.”

### Other

- [ ] I'm interested in working on a PR for this.

AzazelSensei:
## Description

`melos publish --dry-run` runs `dart pub publish --dry-run`. That can exit 65 for a backport / older patch even when a real publish would work (dart-lang/pub#4696). Exec prints FAILED, then we printed "All packages were validated successfully" because the check was `exitCode != 1`.

Only print success when the exec exit code is 0.

Fixes #966

## Type of Change

- [ ] 🗑️ `chore` -- Chore

### DIFF
--- a/packages/melos/lib/src/commands/publish.dart
+++ b/packages/melos/lib/src/commands/publish.dart
@@ -211,7 +211,7 @@ mixin _PublishMixin on _ExecMixin {
       },
     );
 
-    if (exitCode != 1) {
+    if (exitCode == 0) {
       if (!dryRun && gitTagVersion) {
         logger
           ..newLine()
--- a/packages/melos/test/commands/publish_test.dart
+++ b/packages/melos/test/commands/publish_test.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:melos/melos.dart';
 import 'package:melos/src/command_configs/command_configs.dart';
 import 'package:melos/src/command_configs/publish.dart';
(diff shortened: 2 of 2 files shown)


spydon (approved):
Lgtm, thanks for your contribution

PR #1066, opened 2026-08-25
feat(exec): add `--group-logs` to print per-package output once all packages finish

### LINKED ISSUE #134: [melos exec]: Log the outputs in order for each package after everything finished
Using `melos exec` can lead to difficult-to-read logs in concurrent mode, because everything is mixed.

Example:

```
$ melos exec say_hello
[package_a] Hello
[package_b] Hello
[package_a] <Some error>
[package_b] world
```

Since everything is intertwined, tracing what happened in `package_a` is difficult.

We could have `melos exec` print a recap of everything that happened once the command completed, so that we'd have:

```
$ melos exec say_hello
[package_a] Hello
[package_b] Hello
[package_a] <Some error>
[package_b] world

melos exec say_hello completed.

package_b:
Hello
world

package_a:
Hello
<Some error>
```

Ideally, if an `exec` failed in a package, this package should be last in the recap. This way, we do not need to use commands like `more` to see errors if the output is long.

Lyokone:
Resolves #134

## Problem

`melos exec` streams each package's output as it is produced, so with concurrency > 1 the logs interleave and failures can be buried anywhere in the log.

## Change

Adds an opt-in `--group-logs` flag to `melos exec` (and a `groupLogs` option to a script's `exec` block):

- Each package's output is buffered and printed once **all** packages have finished, grouped per package under the same `<package>:` headers `melos analyze` / `melos test` already use.
- Groups print in start order, **except failed packages, which print last**.
- No-op with a single package or `--concurrency 1`; default behaviour unchanged.

Implementation reuses the existing group buffer in `MelosLogger`; the only logger change is a `lastGroups` parameter on `flushGroupBufferIfNeed` that defers the given groups to the end of the flush.

## Verification

`dart analyze` and `dart format` clean. New tests in `test/commands/exec_test.dart` cover grouped output, failures last, no-op at concurrency 1, and the `--fail-fast` / `--order-dependents` combinations.

### DIFF
--- a/docs/commands/exec.mdx
+++ b/docs/commands/exec.mdx
@@ -46,3 +46,30 @@ the script fails in an individual package. Defaults to `false`.
 # Fail fast
 melos exec --fail-fast -- "dart analyze ."
 ```
+
+## --group-logs
+
+When executing a command in multiple packages concurrently, the output of the
+packages is streamed as it is produced, which means the logs of the different
+packages are interleaved and hard to follow.
+
+With `--group-logs`, the output of each package is buffered instead and printed
+once every package has finished, grouped per package and in the order the
+output was produced. Packages in which the command failed are printed last, so
+failures are at the end of the log where they are easy to spot. Defaults to
+`false`.
+
+The flag only has an effect when the command runs in more than one package with
+a concurrency greater than 1. When the command runs in a single package or with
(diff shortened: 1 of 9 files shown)


spydon (approved):
Nice, LGTM

PR #1076, opened 2026-09-09
feat(changed): add a `changed` command that lists the packages changed since a commit or tag

### LINKED ISSUE #48: Feature: melos changed

spydon:
Closes #48.

Adds a `changed` command that lists the local packages that have changed since a given commit or tag.

```bash
melos changed                          # since the latest release tag of each package
melos changed v1.2.0                   # since a tag or commit
melos changed HEAD~5..HEAD             # within a range of commits
melos changed --include-dependents     # plus the packages affected by those changes
```

The command is the `list` command driven by the existing `--diff` filter, with the revision given as an argument instead of an option, so it supports every package filtering option, except `--diff` itself, and every output format of `list`.

### Implementation

- `ChangedCommand` takes an optional `[ref]`. The output flags of `list` moved into a `PackageListOutputOptions` mixin that both commands use, where `list` keeps `--cycles` by overriding `outputKind`. The flag precedence is unchanged.
- `setupPackageFilterParser` takes a `diff` flag so that a command that determines the revision range itself can leave out the `--diff` option.
- `workspaceTag` is passed from the workspace configuration to the diff filter. Without it, a workspace that is versioned in lockstep looks for tags prefixed with the package name that never exist and reports every package as changed.

### Behavior changes of the diff filter

- A package for which no release tag can be found is now always considered changed. The revision range could not be resolved for such a package before, in which case the diff compared the working tree against `HEAD`, so a package that was never released showed up as unchanged. This includes packages without a version in their `pubspec.yaml`, as those can never have a release tag.
- To determine the latest release of a package for a diff, tags of every release type are considered, so a package that is in a prerelease cycle without a prerelease tag is compared against its latest stable tag instead of being considered as never released.
- An invalid revision used to fail with an unhandled `ProcessException` for every package at once, printing a stack trace. The revisions are now verified once, before the packages are filtered, and an invalid one fails with a `MelosException`:

  ```
  Invalid package filters: "v.1.2.0" is not an existing commit, tag or range of commits.
  ```

### Tests

`packages/melos/test/commands/changed_test.dart` builds real git repositories and covers the default of the latest release tag, a workspace without changes, packages that were never released, the prerelease fallback, an explicit revision, a range of commits, `--include-dependents`, an invalid revision, and the argument handling of the command.

### DIFF
--- a/packages/melos/lib/src/common/git.dart
+++ b/packages/melos/lib/src/common/git.dart
@@ -398,7 +398,7 @@ const gitDiffSinceLatestTag = '';
 /// [workingDirectory].
 ///
 /// [diff] is either a single commit or tag, or a range of commits in the git
-/// short hand syntax `<start-commit>..<end-commit>` and
+/// shorthand syntax `<start-commit>..<end-commit>` and
 /// `<start-commit>...<end-commit>`, in which case both ends are checked.
 Future<bool> gitRevisionsExist(
   String diff, {
--- a/packages/melos/lib/src/common/git.dart
+++ b/packages/melos/lib/src/common/git.dart
@@ -617,7 +617,7 @@ Future<String?> _resolveRevisionRange(
     if (_gitVersionRangeShortHandRegExp.hasMatch(revisionRange)) {
       return revisionRange;
     }
-    // If the revision range is not a valid revision range short hand then we
+    // If the revision range is not a valid revision range shorthand then we
     // assume it's a commit or tag and default to the range from that
     // commit/tag to HEAD.
     return '$revisionRange...HEAD';
--- a/packages/melos/lib/src/common/git.dart
+++ b/packages/melos/lib/src/common/git.dart
@@ -282,7 +282,7 @@ Future<bool> gitTagCreate(
 ///       order.
 ///
 ///       Note: If the current version is a prerelease then only prerelease tags
-///       are requested.
+///       are requested, unless a [tagReleaseType] is given.
 ///
 ///       Note: Packages using plain version tags can also have tags prefixed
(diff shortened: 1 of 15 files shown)


packages/melos/lib/src/common/git.dart
```diff
@@ -321,12 +322,14 @@ Future<String?> gitLatestTagForPackage(
...
   // If the current version is a prerelease then only prerelease tags are
   // requested.
-  final tagReleaseType = package.version.isPreRelease
-      ? TagReleaseType.prerelease
-      : TagReleaseType.all;
+  final releaseType =
+      tagReleaseType ??
+      (package.version.isPreRelease
+          ? TagReleaseType.prerelease
+          : TagReleaseType.all);
```
spydon: Can we make this nicer than to have nesting like this?

packages/melos/lib/src/common/git.dart
```diff
@@ -439,19 +480,38 @@ Future<List<GitCommit>> gitCommitsForPackage(
...
   required String? diff,
   required MelosLogger logger,
+  bool workspaceTag = false,
 }) async {
   final revisionRange = await _resolveRevisionRange(
     package,
     diff: diff,
+    workspaceTag: workspaceTag,
+    // Changes are compared against the last release of any type, as a
+    // prerelease is a release like any other here.
```
spydon:
```suggestion
```

PR #1084, opened 2026-09-17
fix(conventional_commit): accept a breaking marker before the scopes

spydon:
## Problem

A `!` placed directly after the type, before the scopes, makes the whole commit unparseable rather than breaking:

| Header | `tryParse` | `isBreakingChange` |
|---|---|---|
| `refactor(scope)!: x` | parsed | `true` |
| `refactor!: x` | parsed | `true` |
| `refactor!(scope): x` | `null` | n/a |
| `chore!(scope): x` | `null` | n/a |

Because `tryParse` returns `null`, the commit is not a conventional commit at all. It is not versionable, it never reaches the changelog, and `melos version` just reports "No packages were found that required versioning", which is indistinguishable from a plain `chore:` commit. A breaking change announced that way is released as if nothing happened.

The mixed behavior is what makes this easy to run into: `refactor!:` is read as breaking, so the `!` looks like it works until a scope is added.

## Fix

The header regex gains an optional `!` between the type and the scopes, and `isBreakingChange` reads that group as well as the existing one after the scopes. `type!(scope)!:` is accepted too.

Nothing else needed changing. Once the commit parses as breaking, `isVersionableCommit`, `includeInChangelog` and `semverReleaseType` already key off `isBreakingChange`, so it bumps (major, which below 1.0.0 means a minor bump) and renders with the `**BREAKING**` prefix under its type heading.

The specification places the `!` after the scopes, so `type(scope)!:` remains the documented form and nothing in the docs changes. Accepting the other placement is deliberate leniency: silently discarding the commit that announces a breaking change is the worse outcome, and this matches how the no-scope form already behaves.

## Tests

- `conventional_commit`: breaking detection, header round-trip, scope parsing and type parsing for `docs!(scope):`, `docs!(scope,dope):` and `docs!(scope)!:`.
- `melos`: `isVersionableCommit` for a non-versionable type marked breaking before the scopes, `semverReleaseType` major for `refactor!(scope):`, and a changelog test asserting `**BREAKING** **REFACTOR**(a): b.` is rendered.

`melos analyze --fatal-infos --fatal-warnings`, `melos format --output none --set-exit-if-changed` and `melos test` all pass locally (555 passed, 1 pre-existing skip).


spydon:
Closing this since the current Conventional Commits 1.0.0 defines that the ! should be put after the scopes.

PR #1085, opened 2026-09-18
feat(config): allow "x-" extension fields so YAML anchors validate

### LINKED ISSUE #571: request:  Support YAML anchors to in Melos YAML schema to limit repetition

### Command

_No response_

### Description

Suppose `melos.yaml` to contain:
```yaml
scripts:
  foo:
    run: echo foo
    exec:
      concurrency: 1
      orderDependents: true
  
  bar:
    run: echo bar
    exec:
      concurrency: 1
      orderDependents: true

  baz:
    run: echo baz
    exec:
      concurrency: 1
      orderDependents: true
```
As you can see, the `exec` configuration is repeated over and over. YAML supports anchors and aliases, so we can instead write this:
```yaml
x-exec: &execConfig
  concurrency: 1
  orderDependents: true

scripts:
  foo:
    run: echo foo
    exec: *execConfig
  
  bar:
    run: echo bar
    exec: *execConfig

  baz:
    run: echo baz
    exec: *execConfig
```
This already works and nothing needs to be done, this YAML feature is supported by the Dart YAML parser.

The only 'issue' is that if the Melos plugin is installed, which turns on Melos YAML schema validiation (or without the plugin, validation can also be turned on) this causes an error:
Screenshot 2023-08-24 at 12 51 05

The name `x-exec` is arbitrary and can be whatever, and it is not expected, hence the error, but this can be fixed. As an example, Docker Compose uses YAML as well for its config and it allows 'extension fields': https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields (this is where the `x-` prefix comes from in my example). Here is how they do it: https://raw.githubusercontent.com/compose-spec/compose-spec/master/schema/compose-spec.json (search for x-), here more about 'pattern properties': https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties

Would it be possible to make it work for Melos as well?

### Reasoning

This would make the schema more flexible and ready for more advanced YAML features, which are already supported anyway.

_No response_

Lyokone:
## Description

Closes #571.

YAML anchors already work in the Melos configuration, because the Dart YAML
parser supports them and Melos ignores keys it does not recognise. The problem
is the key that holds the anchor. It has to live somewhere, and wherever it is
put, `melos.yaml.schema.json` rejects it, so it is reported as an error on a
configuration that Melos itself is perfectly happy with:

```yaml
melos:
  x-analyze: &analyze        # Property x-analyze is not allowed
    command: dart analyze .
    concurrency: 1
    orderDependents: true

  scripts:
    analyze:
      exec: *analyze
```

## Fix

Any key prefixed with `x-` is now an extension field, ignored by Melos and
accepted by the schema. This is the convention the [Compose specification]
uses for the same purpose, and the `x-` prefix in the issue comes from there.

The schema gains `"patternProperties": { "^x-": {} }` everywhere it sets
`additionalProperties: false`. Unknown keys without the prefix are still
rejected, so nothing is given up by allowing it.

`scripts` needed a second change. It is an open map, so the schema was never
the thing rejecting an anchor there, but `Scripts.fromYaml` turned it into a
script: it was listed by `melos run` and, having neither `run` nor `exec`, it
could only fail if picked. It is now skipped, so an anchor can be declared next
to the scripts that alias it:

```yaml
melos:
  scripts:
    x-analyze: &analyze
      exec:
        command: dart analyze .
        concurrency: 1

    analyze: *analyze
```

Melos parses YAML 1.2, which has anchors and aliases but not the YAML 1.1 merge
key (`<<`). An alias therefore replaces a whole value and cannot be merged into
a mapping that sets other keys alongside it. Both forms above are written that
way, and the documentation says so, because `<<` silently does nothing rather
than failing.

Two consequences worth stating plainly, both now documented:

- The example in #571 shares `concurrency` and `orderDependents` between scripts
  running *different* commands. That relied on the Melos 7 shape, where the
  command lived in `run` and `exec` held only options. In Melos 8 the command is
  inside `exec`, so without merge keys an anchor aliased into `exec` has to carry
  the command too, and can only be shared by scripts running the same command.
  What this PR fixes is the blocker the issue actually describes — that the key
  holding the anchor is reported as an error.
- `x-` is reserved in `scripts`, so a script can no longer be named `x-...`. Such
  a script is skipped rather than reported. A warning is not an option, because
  an anchor in `scripts` is by construction indistinguishable from a script; this
  is the same trade the Compose specification makes for the prefix.

## Drive-by: the script schema was still on the Melos 7 shape

Writing the documentation surfaced this: the examples above are valid Melos 8
configuration, and validating them against the schema failed for reasons that
had nothing to do with anchors. `$defs/script` never caught up with 8.0.0.

- `run` and `exec` became mutually exclusive, but the schema still had a branch
  requiring both, and no branch for an object `exec` on its own. So every
  `exec:` script with options was reported as invalid, and `run` + `exec`, which
  Melos rejects with a migration message, was reported as valid.
- An object `exec` carries the command as `exec.command`, which the schema did
  not know about at all.
- `steps`, `private`, `groups` and `stdio` were missing from the script
  properties, and `additionalProperties: false` turned each of them into an
  error.

All four are fixed here, against what `Script.fromYaml` accepts and rejects. The
pre-existing `name` property is left alone: `Script.fromYaml` takes the name from
the map key and never reads `name:`, but it is documented in
`docs/configuration/scripts.mdx` and removing it is out of scope here.

`stdio` is constrained rather than just added, since Melos rejects
`stdio: inherit` alongside `exec` or `steps`. The steps-only branch of
`$defs/script` is dropped: the `steps` branch added to the inner `anyOf` covers
it with the full property set, so it could never be the only match.

## Known limitation, not addressed here

The schema's top level is still the Melos 6 `melos.yaml` document: it requires
`name` and `packages`, and knows nothing about `workspace:` or the `melos:`
block, so it does not validate a Melos 8 `pubspec.yaml` as-is. Everything above
applies to the `melos:` block's contents, which is what editor integrations map.
`docs/ide-support.mdx` still scopes the schema to "Melos v6 and lower" and points
at the copy in `blaugold/melos-code`. Bringing the top level up to Melos 8 is a
larger change than this one and I have left it alone — happy to follow up if you
want it.

## Tests

`Scripts.fromYaml` ignoring `x-` keys is covered in `workspace_config_test.dart`.

The schema has no test harness in the repository, so I validated it out of tree
with `jsonschema`, against Melos' own configuration, both documentation
examples, and a script using `steps`, `private`, `groups` and `stdio`; and
asserted that `run` + `exec`, an object `exec` without `command`, an unknown
top-level key and an unknown script key are all still rejected.

## Checklist

- [x] `dart analyze --fatal-infos --fatal-warnings` is clean
- [x] `melos format --output none --set-exit-if-changed` is clean
- [x] Tests pass, with the same 28 failures before and after the change

Those 28 are environmental: the integration tests shell out to a `melos` on the
`PATH`, and I have no global installation activated, so they fail with
`ProcessException: No such file or directory`. The same 28 fail on `main` at
5fcc986 with the change stashed.

## Type of Change

- [ ] 🗑️ `chore` -- Chore

[compose specification]: https://docs.docker.com/reference/compose-file/extension/


spydon (approved):
Lgtm!

